PR #7, opened 2022-09-29
Refactors Condition with 2 seperate overloaded run methods

spokenbird:
- Run now takes just a Submission or a Submission and a String (usually UUID)

### DIFF
--- a/lib/src/main/java/formflow/library/config/submission/Condition.java
+++ b/lib/src/main/java/formflow/library/config/submission/Condition.java
@@ -17,5 +17,11 @@ public interface Condition {
    * @param submission submission object the condition is associated with, not null
    * @return true if the condition check passes, else false
    */
-  public Boolean run(Submission submission);
+  public default Boolean run(Submission submission) {
+    return null;
+  }
+
+  public default Boolean run(Submission submission, String uuid) {
+    return null;
+  }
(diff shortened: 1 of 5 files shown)


lib/src/main/java/formflow/library/config/submission/Condition.java
```diff
@@ -17,5 +17,11 @@ public interface Condition {
...
    * @return true if the condition check passes, else false
    */
-  public Boolean run(Submission submission);
+  public default Boolean run(Submission submission) {
+    return null;
+  }
+
+  public default Boolean run(Submission submission, String uuid) {
+    return null;
+  }
```
bseeger, line 26: Great idea to have these as implemented functions we can override. We could copy this paradigm for the Actions as well.  

Personally I'd rather they return `false` instead of `null`, or, better yet, we could have them throw an error like so: 

`throw UnsupportedOperationException("Not implemented");`

Or something like that to indicate to the implementors that the condition didn't just fail, it's not setup at all in the first place and needs their attention if they plan to use that method in their subclass.

spokenbird: Great ideas, I think the UnsupportedOperationException would be the way to go :)

PR #10, opened 2022-10-04
Adds a message properties file. 

bseeger:
Adds a messages properties file named `messages-form-flow.properties` to this library.  The file name must not conflict with what's in the template starter app, so I gave it this name. 

This pulls in the general properties for the site - though we may want to audit the fragments and pull in any property that's found in those.


coltborg (approved):
Paired, lgtm

PR #12, opened 2022-10-11
WIP: Removes use of Hibernate Types library and uses a converter instead

bseeger:
This is a proof of concept for pulling out the Hibernate Types library and doing the conversion ourselves. 

There might be other ways to do this as well with just annotations and no custom code.

### DIFF
--- a/lib/src/main/java/formflow/library/utils/HashMapConverter.java
+++ b/lib/src/main/java/formflow/library/utils/HashMapConverter.java
@@ -1,0 +1,41 @@
+package formflow.library.utils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import javax.persistence.AttributeConverter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class HashMapConverter implements AttributeConverter<Map<String, Object>, String> {
+
+    @Override
+    public String convertToDatabaseColumn(Map<String, Object> attribute) {
+        String stringInfo = null;
+
+        ObjectMapper objectMapper = new ObjectMapper();
+
+        try {
+            stringInfo = objectMapper.writeValueAsString(attribute);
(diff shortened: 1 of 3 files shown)


lib/src/main/java/formflow/library/utils/HashMapConverter.java
```diff
@@ -0,0 +1,41 @@
+package formflow.library.utils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
```
spokenbird, line 3: Is Jackson only being used for the exception in this case?

bseeger: The ObjectMapper is from Jackson as well.

lib/src/main/java/formflow/library/utils/HashMapConverter.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import javax.persistence.AttributeConverter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
```
bseeger, line 10: I can add some comments tomorrow - I borrowed some of this idea from a website which I can credit as well.

bseeger:
I made a mistake when making this PR, which was that I didn't have my form-flow-starter-library pointing to my dev form-flow library.  This PR currently doesn't work as is. Sorry about that!  I'm working on getting it to work now. :)

bseeger:
I ended up getting this to work if I update this line in the form-flow-starter-app's `application.yaml` file: 

```
  datasource:
    url: jdbc:postgresql://localhost:5432/starter-app?stringtype=unspecified
```

I added the `stringtype=unspecified` which tells postgres to relax its type restrictions some.  Not actually sure if this is a good idea though.

coltborg:
After reviewing in stand up, we're going to close this PR and keep `hibernate-types` for now.

PR #13, opened 2022-10-14
WIP: Removes slashes from View's in controller.

bseeger:
Also this tweaks the name of the GPR library name, as this is really just a tests. 

This will need more work - it was a quick pass at testing out removing the `/`'s from the paths. It appears to work with a few issues that need to be sorted out. Namely, I get a few pages in (yeah!) and then hit a wonky path it creates, which is probably due to me missing some logic in removing the slashes. Example failing path: `http://localhost:8080/ubi/ubi/languagePreference/navigation`

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -87,7 +87,7 @@ publishing {
             // eventually this should probably be org.codeforamerica.form-flow ...?
             group = "form-flow"
             artifactId = "form-flow-library"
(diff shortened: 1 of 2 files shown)


bseeger:
This appears to fix the docker issue and still work locally. 

Resource: https://stackoverflow.com/questions/57221353/thymeleaf-fails-to-resolve-templates-when-running-inside-docker-container

PR #17, opened 2022-10-20
Reformats the code based on the google stylesheet 

bseeger:
Includes Ben's change for extended hard wrap at 130 characters

I applied the reformatting to the java and html files, but stayed away from the stylesheets and sql file(s).


src/main/java/formflow/library/config/TemplateManager.java
```diff
@@ -10,7 +10,6 @@
 import java.lang.reflect.Constructor;
 import java.util.HashMap;
 import java.util.List;
-import java.util.stream.Collectors;
```
bseeger, line 13: I took this out as it's unused.

coltborg: Ohh we can tell [IntelliJ to remove unused imports on the fly](https://intellij-support.jetbrains.com/hc/en-us/community/posts/206915775-How-to-enable-remove-unused-imports-#community_comment_207047099).

coltborg (approved):
lgtm

PR #20, opened 2022-10-21
Makes static methods in Submission object methods instead

bseeger:
This updates the Submission class and any spot that referenced them from the java and html files.


src/main/java/formflow/library/data/Submission.java
```diff
@@ -60,6 +59,11 @@ public class Submission {
   @Column(name = "submitted_at")
   private Date submittedAt;
 
+  public Submission() {
+    inputData = new HashMap<>();
```
bseeger, line 63: This setup gets rid of the error we were seeing around line 46 above when we compiled.  Line 46 was creating a HashMap that would technically get overwritten when the setter for it was called, as it will pass in a new HashMap each time.  If we don't populated it with a HashMap to start with though, we will hit other errors later on in the code that assume we always have something set here.

PR #50, opened 2022-11-28
Add Maven publish support

coltborg:


build.gradle
```diff
@@ -85,22 +85,80 @@ test {
...
         maven {
-            name = "GitHubPackages"
-            url = uri("https://maven.pkg.github.com/codeforamerica/form-flow")
+            name = "MavenCentral"
+            url = "https://s01.oss.sonatype.org/content/repositories/snapshots"
             credentials {
-                username = props.getProperty("USERNAME") ?: System.getenv("USERNAME")
-                password = props.getProperty("TOKEN") ?: System.getenv("TOKEN")
+                username = localEnv.getProperty("MAVEN_USERNAME")
+                password = localEnv.getProperty("MAVEN_PASSWORD")
```
bseeger, line 101: Just noting for now, as we don't need to change this yet:  I think we still need the ability to pull the information from the environment, as Aptible will need that.

bseeger: Also noting that we'll need to update the sample env file once this settles out.

PR #52, opened 2022-11-30
Delete uploaded file

coltborg:


src/main/java/formflow/library/UploadController.java
```diff
@@ -168,17 +168,19 @@ RedirectView delete(
               // - remove things after .ext
               // - have '-thumbnail' before the .txt
               // - add txt as extension
+              // TODO - maybe this should be replaceLast, so if there are other '.' in the string we don't run into issues?
```
coltborg, line 171: @bseeger is there a `replaceLast`? I only see `replace` and `replaceAll` in my autocomplete?

coltborg: We also control the filename, so if there's another `.` it would be because we add it to the cloud filename.

coltborg: All this being said, the current regex here doesn't work 😢 

Screenshot 2022-12-01 at 5 45 37 PM

coltborg: I had to use `.replaceFirst` as it would allow regex. `.replace()` does not accept regex to find a target. Only `.replaceFirst` and `.replaceAll`.

PR #61, opened 2022-12-06
Fix deploy github action to not deploy on PR's

coltborg:
Use same workflow_run as deploy in starter app


bseeger (approved):
:) given that deploy didn't run on this PR and tests did, this looks good to me.

PR #72, opened 2022-12-13
Fix tif isues

bencalegari:
Co-authored-by: Bethany Seeger <bseeger@codeforamerica.org>

[#183995116]


bencalegari:
@bseeger take two! moved this to the library, merge if it looks good!

PR #74, opened 2022-12-14
Adds `reveal` fragment

bseeger:
This adds a `reveal` fragment loosely based on what was in Shiba.  

Also adds icons necessary for the Document Recommendations page.


bseeger:
used by: https://github.com/codeforamerica/form-flow-starter-app/pull/40

src/main/resources/META-INF/resources/webjars/form-flow/0.0.1/css/libraryStyles.css
```diff
@@ -393,6 +393,10 @@ body,
   color: #FFFFFF;
 }
 
+.button--other {
```
bseeger, line 396: used by `Document Recommendations` page.  Not sure if this is a good name or not, but needed to update the margin to get it to look correct.

PR #84, opened 2022-12-21
Adds DevController and icons end point

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/184048149

Adds new DevController under a new `dev` profile.  When using the `dev` profile, and icons endpoint will be available at `/dev/icons`. 

Updated readme to include instructions on how to switch to this profile and access the endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Table of Contents
     * [Set up jenv to manage your jdk versions](#set-up-jenv-to-manage-your-jdk-versions)
     * [Gradle](#gradle)
         * [Build Web/Fat Jar](#build-webfat-jar)
+    * [Spring Profile: `dev`](#spring-profile-dev)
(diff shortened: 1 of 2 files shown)


bencalegari:
@bseeger I updated DeepSource so it should no longer fail builds when it finds things it doesn't like, but in this case I'd recommend looking at [what it reported about using `@RequestMapping` rather than a request method specific approach.](https://deepsource.io/gh/codeforamerica/form-flow/run/2b8a6fa9-3f69-4679-bc01-2485d9cc7dfe/java/JAVA-S1065?listindex=0)

bseeger:
Thanks, @bencalegari. I guess I can try to limit it a specific type of call, but it's complaining about something that's in a different scope than I am.  I have this `@RequestMapping` at the _class_ level, not function level.  I'll look into seeing if I can protect it more later, though.

coltborg (approved):
LG, aside from whatever Deepsource doesn't like.

PR #91, opened 2023-01-03
Dropzone configuration settings and some max file fixes

bseeger:
fixes issues in: https://www.pivotaltracker.com/story/show/184138449
some work for: https://www.pivotaltracker.com/story/show/183559454

When the maximum number of files was hit, dropzone and the document upload page were not doing the correct thing. 

Now an error message box is displayed when the max file number is reached.   When returning to the page, the max file number may not be exceeded either.

### DIFF
--- a/src/main/resources/templates/fragments/fileUploader.html
+++ b/src/main/resources/templates/fragments/fileUploader.html
@@ -71,13 +71,12 @@
     window['userFileIds' + [[${inputName}]]] = [];
     window['cancelledFiles' + [[${inputName}]]] = [];
     var userFiles = [[${session.userFiles}]]
-    // TODO: move to more global setting
-    // - either coming from fragment parameter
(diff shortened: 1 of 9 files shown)


src/main/resources/templates/fragments/fileUploader.html
```diff
@@ -386,7 +395,8 @@
               name: doc.originalFilename,
               size: doc.filesize,
               type: doc.type,
-              id: parseInt(uploadedDocWithThumbnail[0])
+              id: parseInt(uploadedDocWithThumbnail[0]),
+              accepted: true
```
bseeger, line 399: This fixes the issue of dropzone not noticing that we hit the max file limit if we had the max files, clicked continue and then went back.  

Dropzone has the concept of "accepted" files, so if they were not accepted, they didn't count towards maxFiles.

PR #96, opened 2023-01-06
DO NOT MERGE (yet): Changes file id to UUID versus just an int

bseeger:
addresses: https://www.pivotaltracker.com/story/show/184056296

Please note this changes the database, so if you test this, be prepared for that. 

This changes the file id in `user_files` to be a uuid instead of an integer.


bseeger:
Tests need to be fixed... so still a WIP

src/main/resources/db/migration/V4__user_files_change_primary_key_type.sql
```diff
@@ -0,0 +1,7 @@
+-- This will make the original file_id a column a uuid column
+create extension if not exists "uuid-ossp";
```
bseeger, line 2: In theory postgreSQL vs 14 (which I think we use) should have this function that we could use w/o creating an extension: 
```
alter table user_files add column new_id uuid DEFAULT gen_random_uuid();
```

I've tried (and failed) to get that variant to work with tests.

bseeger: From what I can see flyway is trying to instantiate a bean of the db stuff, and that's causing migrations to run, and fail.  It's not really the tests themselves, I think.

bseeger:
Merging this will cause the snapshot to be updated with this change and it will affect the dev/demo databases. That's fine, but we need to make sure that folks understand about making sure they pull the changes in the starter app as well, or else the system will act weird for them.  Not a big deal, just need to coordinate a little to make sure we are all ready for the DB changes this brings.

PR #105, opened 2023-01-18
Protect input names with inherited class from library

spokenbird:
This PR adds a class `FlowInputs` which all flow input classes will inherit from. Example `public class Ubi extends FlowInputs`. 

The parent class now holds the _csrf token field.

Having all flow inputs classes extend this class will also enforce that input fields are always declared in a corresponding flow inputs class that extends the FlowInputs class. 

One unintended but seemingly unavoidable change this PR introduces, is the need to camel case input names so that they can be used as fields in a flow inputs class (i.e. a field name cannot be `doc-upload-files` it has to be `docUploadFiles`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,8 +328,8 @@ extend the class `FlowInputs` which provides CSRF functionality for security.
 Also note that for single value inputs the type when defining the input is String. However, for
 input types that can contain more than one value, the type is ArrayList<String>.
 
-When naming your inputs in your templates, make sure that you use camel case so that the given input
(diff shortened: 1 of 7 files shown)


bseeger:
We've had to have camel case for the Ubi.java file for a while because of validation, so I think it's just that the docUpload stuff wasn't up to date with that. Now that validation happens for that form, too, it's forced us to fix something that'd just been overlooked up til now.   Hope that makes some sense.

PR #113, opened 2023-01-24
Document input types and their parameters in README

spokenbird:
[#184078352]

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,7 +639,7 @@ Text inputs are used to gather text input from the user. They can be used to gat
 names, addresses, and emails, etc.
 
 Text inputs have an optional placeholder parameter which can be used to display a placeholder within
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -600,7 +600,321 @@ into your template to quickly see a preview and names of icons:
...
+```
+
+Note that the `checkboxInSet` fragment is used to provide multiple options within
+a `checkboxFieldset` fragment. Also note that the input name fo the `checkboxFieldset` and
+the `checkboxInSet`
+are the same. This is how the fieldset and internal checkbox options are grouped into a single
+multiple checkbox
+input.
+
+Convenience
```
bencalegari, line 796: @spokenbird Was this meant to be replaced with a reference to the livetemplate?

PR #122, opened 2023-01-31
Update number input to use inputmode="number" and pattern paradigm

spokenbird:
- Default pattern allows zero, whole numbers, or decimals with up to two decimal points
- Negative numbers are not allowed
- Pattern can be changed by passing a different pattern
- Title can be changed by passing a different title


coltborg (approved):
Looks good. Tested locally and worked as expected.

PR #128, opened 2023-02-04
Fix LiveTemplates.xml.   

enyia21:
Some extra code was added to LiveTemplates that broke it.  This commit fixes that


bseeger (approved):
Good catch -- changes look good.

coltborg:
Is this still good to go?

bseeger:
The same changes for this commit were brought in via another PR, so this one is no longer necessary.

PR #129, opened 2023-02-04
Submit starter app add icons to success

enyia21:


src/main/resources/templates/fragments/icons.html
```diff
@@ -1967,6 +1967,84 @@ <h1>Icons</h1>
...
+            <path d="M65.8578 58.5313L71.0777 64.3581L81.8549 53.3481" stroke="black"
+                  stroke-width="4" stroke-linecap="round" stroke-linejoin="round"/>
+          </g>
+          <defs>
+            <clipPath id="clip0_138_424">
+              <rect width="100" height="75" fill="white"/>
+            </clipPath>
+          </defs>
+        </svg>
+
```
bseeger, line 2045: This extra line is not needed.

src/main/resources/templates/fragments/icons.html
```diff
@@ -1967,6 +1967,84 @@ <h1>Icons</h1>
...
+        <svg th:fragment="downloadDocumentsIcon" aria-hidden="true" width="30" height="35"
+             viewBox="0 0 30 35" fill="none" xmlns="http://www.w3.org/2000/svg">
+          <path
+              d="M26 12.6471L18.5714 12.6471V1L7.42857 1L7.42857 12.6471L0 12.6471L13 26.2353L26 12.6471ZM0 30.1177L0 34H26V30.1177L0 30.1177Z"
+              fill="black"/>
+        </svg>
+      </td>
+    </tr>
+    <tr>
+      <td>successComputerPrimaryIcon</td>
```
bseeger, line 2000: We haven't usually put the word "Icon" in the name of the item, just the description of what it is.  Is it just a computer?

src/main/resources/templates/fragments/icons.html
```diff
@@ -1967,6 +1967,84 @@ <h1>Icons</h1>
         </svg>
       </td>
     </tr>
+    <tr>
+      <td>submitAdditionalIcons</td>
```
bseeger, line 1971: We haven't put the word "Icons" in the names so far.   What is this icon of?  The name doesn't make it clear -- should be what it is, not what it's for.

src/main/resources/templates/fragments/icons.html
bseeger: Maybe this could be  `computerWithCheckmark`?

src/main/resources/templates/fragments/icons.html
bseeger: Also, this is a file, so it's probably better to name it "file" or something relating to what it looks like, not what we are using it for in the application.  That way others can get an idea of what it looks like and use it as they please.

src/main/resources/templates/fragments/icons.html
bseeger: Actually, I take this back.  We have a varied practice now with the word "Icon", so what you did is not inconsistent with what's in the file, but I think we should take the word Icon out from all spots.

src/main/resources/templates/fragments/icons.html
bseeger: Actually, I take this back about the word "Icon". We have a varied practice now with the word "Icon", so what you did is not inconsistent with what's in the file, but I think we should take the word Icon out from all spots.

bseeger:
This looks good overall with a few small things that need to be addressed. Good work!

bseeger:
This looks good overall with a few small changes needed.  Good work!

src/main/resources/templates/fragments/icons.html
enyia21: Updated to: smallFile

src/main/resources/templates/fragments/icons.html
enyia21: Updated to: computerWithCheckmark

bseeger (approved):
Looks good!

PR #131, opened 2023-02-07
fix money validator - previously disallowed single digit numbers

lkemperman-cfa:


bseeger (approved):
LGTM! Thanks for updating this regex.

PR #134, opened 2023-02-09
Create new message collection for each key

sree-cfa:


src/test/java/formflow/library/utilities/AbstractMockMvcTest.java
```diff
@@ -331,6 +331,15 @@ protected void postExpectingFailureAndAssertErrorsDisplayForThatInput(String pag
     assertInputHasErrors(pageName, inputName, errorMessages);
   }
 
+  protected void postExpectingFailureAndAssertInputErrorMessages(String pageName,
+      Map<String, String> inputParams,
+      Map<String, List<String>> expectedErrorMessages) throws Exception {
+//    postExpectingFailure(pageName, inputParams);
```
sree-cfa, line 337: How are there error messages without this post?

bseeger: I think this test really operates via the `assertInputHasErrors` call: https://github.com/codeforamerica/form-flow/pull/134/files#diff-63e6c79523ceedc33846c5a80d3fb2613b1124f70e5374c7338e4ae3c19ef8eaR398

bseeger (approved):
This looks good to me. Thanks, @sree-cfa, for fixing this and adding more to the tests to catch future regressions. 

We will look further into the question you have on the test out side of this ticket, if that sounds good.

bseeger:
The fix looks great, but if you wouldn't mind holding off on merging, we can figure out why the tests fail even though you haven't posted the data.

src/test/java/formflow/library/utilities/AbstractMockMvcTest.java
```diff
@@ -395,7 +404,14 @@ protected void assertInputHasErrors(String pageName, String inputName, Integer n
...
-    assertTrue(errorMessages.containsAll(page.getInputErrors(inputName).stream().map(Element::ownText).toList()));
+    // make sure there are errors returned
+    assertFalse(page.getInputErrors(inputName).isEmpty(), "Expected errors on page, but there were none");
+    // assert equal amount
+    assertEquals(errorMessages.size(), page.getInputErrors(inputName).size(),
+        "The page has a different number of errors than expected");
+    // now check if they match the expected ones.
+    assertTrue(errorMessages.containsAll(page.getInputErrors(inputName).stream().map(Element::ownText).toList()),
+        "The error messages expected do match what was returned");
   }
```
bseeger, line 415: This function needed more. There were two issues with the current check: 
 * The logic of `List.containsAll(empty List)` will return true, since a set always contains an empty set.  So the logic here was that if the page hadn't POSTed and there were no errors, this function thought things were fine, since no errors is a subset of the errors list. 
 * Also this didn't take into account if the number of errors returned by the page were less than the ones expected. It just makes sure that the error messages list contains all that the page returned.   If the page returned less errors than expected, it still thought things were fine. 

Adding the other two asserts should account for these scenarios.

src/test/java/formflow/library/utilities/AbstractMockMvcTest.java
bseeger: If the change to the assert function below makes sense, I think we can uncomment this line.

PR #136, opened 2023-02-13
Adds demo banner to header.

bseeger:
Updates Demo Banner to trigger on a spring profile "demo"
Addresses: https://www.pivotaltracker.com/story/show/184420950

Revamps the demo banner fragment to display the banner if a certain profile is active, like "demo". The user can no pass in the profile they want it displayed on.

### DIFF
--- a/src/main/resources/templates/fragments/head.html
+++ b/src/main/resources/templates/fragments/head.html
@@ -6,7 +6,5 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="format-detection" content="telephone=no">
   <title th:text="${title}"></title>
-  <th:block
-      th:replace="fragments/demoBanner :: demoBanner(demoText=#{demo.banner-text}, profile='demo')"/>
 </head>
 </html>
--- a/src/main/resources/META-INF/resources/webjars/form-flow/0.0.1/css/libraryStyles.css
+++ b/src/main/resources/META-INF/resources/webjars/form-flow/0.0.1/css/libraryStyles.css
@@ -583,3 +583,7 @@ body,
 }
 
 /* end reveal */
+
+.demo-banner {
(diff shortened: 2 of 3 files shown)


src/main/resources/templates/fragments/head.html
```diff
@@ -6,5 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="format-detection" content="telephone=no">
   <title th:text="${title}"></title>
+  <th:block
```
coltborg, line 9: Actually I don't think it's a good idea to put [rendered content inside the `<head>`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML)

> The head of an HTML document is the part that is _not displayed_ in the web browser when the page is loaded.

But your idea to put it in the `head` fragment is still a good one.

My thought it we could do something like this:

```html
<th:block th:fragment="head" th:assert="${!#strings.isEmpty(title)}">
  <head>...</head>
  <th:block th:replace="demoBannter" />
</th:block>
```

We can pair on this if you'd like. Or welcome to pushback.

coltborg: Ugggghhh actually it's more complicated, as all valid HTML that [visually appears on the screen needs to be within the `<body>` tag](https://teamtreehouse.com/community/why-is-it-not-valid-to-place-elements-outside-when-body-element-it-is-used-as-the-wrapper).

coltborg: Ok, next best solution: we add `demoBanner` into the `toolbar` fragment (either above or below). I assume every screen/static page will use the `toolbar` as that's where our logo is.

coltborg (approved):
Tested locally by switching the target profile to `dev`. LGTM :shipit:

PR #137, opened 2023-02-14
add additional js function to formFlowDropZone.js to disable if no fi…

lkemperman-cfa:
…les. could be consolidated into same function if we also pass the class


bseeger (approved):
This looks good to me. Makes me think we should change the first function's name to be more generic, but that's outside the scope of this.

PR #138, opened 2023-02-15
Using 'disabled' property

sree-cfa:
I _think_ this is more accessible than class name

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

Elements with the `disabled` attribute are skipped over by screen readers. I think the class itself is not enough indication for other readers to identify it as "disabled"


sree-cfa:
Going to close this for now and explore other options

PR #140, opened 2023-02-17
Show appropriate error when password-protected PDF is uploaded

spokenbird:
This adds the PDFBox library as a dependency and uses it in the UploadController to attempt generating a PDF document from uploaded PDFs. If the upload fails due to password protection, the controller will now respond with ResponseEntity that includes the [422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) status code, and an appropriate error message. The same will occur if an IO error occurs when trying to generate the PDF.

The PR also adds a password-protected PDF upload to the UploadJourneyTest to check for the correct error message and error state.

[#184317984]


build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
     implementation 'net.coobird:thumbnailator:0.4.19'
     implementation 'org.webjars.npm:dropzone:5.9.3'
     implementation 'com.amazonaws:aws-java-sdk-s3:1.12.364'
+    implementation 'org.apache.pdfbox:pdfbox:2.0.27'
```
bseeger, line 69: I'm noticing we've gotten very specific in our release numbers. Maybe this should be locked to major & minor but not patch?  Maybe this could be "2.0.+"?

src/main/java/formflow/library/UploadController.java
```diff
@@ -61,6 +65,21 @@ public ResponseEntity<?> upload(
...
       }
       String fileExtension = Files.getFileExtension(Objects.requireNonNull(file.getOriginalFilename()));
+
+      if (fileExtension.equals("pdf")) {
+        try (PDDocument pdfFile = PDDocument.load(file.getInputStream())) {
+        } catch (InvalidPasswordException e) {
+          // TODO update when we add internationalization to use locale for message source
+          String message = localeLibraryConfiguration.messageSource()
+              .getMessage("upload-documents.error-password-protected", null, null);
+          return new ResponseEntity<>(message, HttpStatus.UNPROCESSABLE_ENTITY);
```
bseeger, line 75: I wonder if a 415 is more appropriate - it's an unsupported media type, though the type itself is supported.

src/test/java/formflow/library/upload/UploadJourneyTests.java
```diff
@@ -24,47 +24,58 @@ public void setUp() throws IOException {
...
         "This file is too large and cannot be uploaded (max size: " + maxFileSize + " MB)");
     testPage.clickLink("remove");
-    Assertions.assertThat(testPage.findElementTextById("number-of-uploaded-files-uploadTest")).isEqualTo("0 files added");
+    assertThat(testPage.findElementTextById("number-of-uploaded-files-uploadTest")).isEqualTo("0 files added");
+
+    // Upload a password-protected file and assert the correct error shows
+    uploadPasswordProtectedPdf("uploadTest");
+    assertThat(testPage.findElementsByClass("text--error").get(0).getText())
+        .isEqualTo(
+            "We are unable to process password-protected PDF files. Please remove the password and try again or try another PDF file.");
```
bseeger, line 52: Do you know if we are able to easily pull in the `messages-form-flow.properties` file here?  It will make future developer's lives easier if someone tweaks those messages.   They/we won't have to find all these satellite spots to change.

src/main/java/formflow/library/UploadController.java
bseeger: Never mind, 422 is fine.

src/test/java/formflow/library/upload/UploadJourneyTests.java
bseeger: If we can do this, maybe it's worth doing in a separate ticket.

PR #142, opened 2023-02-21
Adding id to continue button link

sree-cfa:
Adding id to continue button link for easier selection and tracking.

Button/Link text can change based on locale - `id` can be used to track it across locales


coltborg (approved):
LGTM;

I hope folks won't add multiple continue buttons on a screen, which would cause HTML errors. But I think that's too small of an edge case to address right now.

sree-cfa:
> LGTM;
> 
> I hope folks won't add multiple continue buttons on a screen, which would cause HTML errors. But I think that's too small of an edge case to address right now.

Yeah, I was thinking about that too - axe tests would catch duplicate ids and sounds like we'll have that at some point?

PR #143, opened 2023-02-23
Add support for beforeSave on non-subflows

sree-cfa:


sree-cfa:
Where would be a good place to write tests? form-flow and/or form-flow-starter-app?

src/main/java/formflow/library/data/Submission.java
```diff
@@ -97,7 +97,7 @@ public Submission() {
    */
   public Map<String, Object> getSubflowEntryByUuid(String subflowName, String uuid) {
     if (inputData.containsKey(subflowName)) {
-      ArrayList<Map<String, Object>> subflow = (ArrayList<Map<String, Object>>) inputData.get(subflowName);
+      List<Map<String, Object>> subflow = (List<Map<String, Object>>) inputData.get(subflowName);
```
sree-cfa, line 100: Was getting an error when trying to cast Immutable List to ArrayList

README.md
```diff
@@ -297,6 +298,35 @@ The `inputData` field is a JSON object that stores data from the user's input as
...
+} 
+```
+
+### beforeSave
+
+An action can be added to a page in the flow-config to update the submission data before saving to the database.
+
+```yaml
+expense:
+  beforeSave: actions.CalculateBeforeSave
```
bseeger, line 325: I like how this looks, but really it needs to be a full path to the class: https://github.com/codeforamerica/form-flow-starter-app/blob/main/src/main/resources/flows-config.yaml#L49

Maybe this could be `java.path.to.actions.CalculateBeforeSave`.  This could link to the starter app example in the link here. 

I'd love to figure out how to make this just the class name, but couldn't figure out how to reference the `actionsPath` created earlier in the yaml file: https://github.com/codeforamerica/form-flow-starter-app/blob/main/src/main/resources/flows-config.yaml#L3

src/main/java/formflow/library/ScreenController.java
```diff
@@ -131,6 +131,7 @@ ModelAndView postScreen(
     // if there's already a session
     if (submission.getId() != null) {
       submission.mergeFormDataWithSubmissionData(formSubmission);
+      handleBeforeSaveAction(getScreenConfig(flow, screen), submission);
```
bseeger, line 134: I _think_ you want this outside of whether or not a submission already exists.  This way, the action occurs regardless of whether or not previous data existed.  

I was thinking that maybe this could be done before this if/else statement, though I can see this is tricky with the merging of form data needing to happen first (if submission data already exists).  Given that, I do think the `handleBeforeSaveAction` may need to be called in the else as well.  Esp. if the before save action only depends on currently entered data - we'd still want the action to happen, even if no previous data existed.

src/test/java/formflow/library/submissions/actions/FormatDateBeforeSave.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+
+@SuppressWarnings("unused")
+public class FormatDateBeforeSave implements Action {
+  public void run(Submission submission) {
+    List<String> dateComponents = new ArrayList<>(3);
+    dateComponents.add((String) submission.getInputData().get("dateMonth"));
+    dateComponents.add((String) submission.getInputData().get("dateDay"));
+    dateComponents.add((String) submission.getInputData().get("dateYear"));
+    submission.getInputData().put("formattedDate", String.join("/", dateComponents));
```
bseeger, line 15: :) When I left my comment above about where to run the save from, I thought of this scenario!

src/test/java/formflow/library/submissions/actions/CalculateTotalBeforeSave.java
```diff
@@ -0,0 +1,21 @@
...
+@SuppressWarnings("unused")
+public class CalculateTotalBeforeSave implements Action {
+
+  public void run(Submission submission, String _id) {
+    List<Map<String, Object>> subflow = (List<Map<String, Object>>) submission.getInputData()
+        .get("income");
+    var totalIncome = subflow.stream()
+        .map(e -> Integer.parseInt((String) e.get("textInput")))
+        .reduce(Integer::sum)
+        .get();
```
bseeger, line 17: We do allow folks to enter income having a decimal value (example: $10.50), which would be lost here with an Integer. This should probably be a Float/Double instead.   Or we need away to allow for both scenarios, where truncating the fractional part of the value is okay?

How much do you care about it being an Integer versus Double?    I know the fractional value doesn't amount to much when we are talking about hundreds or thousands of dollars - we just need to align on this.

src/main/java/formflow/library/ScreenController.java
bseeger: This is subtle, too, as a test would miss this unless the action were on the first screen of inputs.

src/main/java/formflow/library/data/Submission.java
bseeger: That's strange that we haven't gotten that before.   It's probably better being a List anyways.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -131,6 +131,7 @@ ModelAndView postScreen(
...
       saveToRepository(submission);
```
bseeger, line 135: Thinking bigger picture here. We allow the user to do whatever they want with the data before they save it.  This is _after_ we validate the data. So we know the initial data is good, but you could introduce all sorts of invalid data via this hook.   A little bit of a yellow flag for me.  But presumably the app writer knows what they are doing. 

Also, in the case of the date action, it allows you to set the date, but not actually validate that it's a good / real one, unless the action hook does it itself.  The fields have all been validated so far, so hopefully that's good and solid, but you have no sense if the aggregated Date field is solid. 

It's making me think that we need a pre-validation hook to manipulate the data. Then we could have (in the FlowInputs Class (UBI.java)) separate fields -- day, month, year, Date. Then Date could be validated as well.  This could solve some issues we are facing with the custom validations. 

If we have this pre-validation action hook, we could check multiple fields and set flags in the object. Then during validation, if the flag doesn't check out, we trigger an error.

src/test/java/formflow/library/submissions/actions/CalculateTotalBeforeSave.java
sree-cfa: Hm, I just used `int` to have something really basic for testing. Ive seen it both ways for real income inputs so no strong preference here.

src/main/java/formflow/library/ScreenController.java
sree-cfa: Yeah that makes sense

src/main/java/formflow/library/ScreenController.java
```diff
@@ -131,13 +131,13 @@ ModelAndView postScreen(
...
-      saveToRepository(submission);
     } else {
       submission.setFlow(flow);
       submission.setInputData(formSubmission.getFormData());
-      saveToRepository(submission);
-      httpSession.setAttribute("id", submission.getId());
     }
+    handleBeforeSaveAction(getScreenConfig(flow, screen), submission);
+    saveToRepository(submission);
+    httpSession.setAttribute("id", submission.getId());
```
sree-cfa, line 140: This is probably already set most of the time, but it seems ok to set it again anyways?

README.md
sree-cfa: Agreed! Not sure how to do that rn tho - will keep it in mind

src/test/java/formflow/library/submissions/actions/CalculateTotalBeforeSave.java
bseeger: Yeah, no strong preference here either. Since we've assumed a fractional value so far, would you mind changing this to a Double?

src/main/java/formflow/library/ScreenController.java
bseeger: Yeah, I don't think this will be an issue.

bseeger (approved):
Looks good! Thanks for the changes.

PR #146, opened 2023-02-28
add background image path from webjar for list--checkmark items

lkemperman-cfa:


sree-cfa:
Are we able to update the paths in the honeycrisp.css itself?

lkemperman-cfa:
> Are we able to update the paths in the honeycrisp.css itself?

Technically yes - but I was trying to isolate the honeycrisp asset pipeline from any overrides that we are doing to make these assets available in the java environment. It seems like that is the pattern that has been used so far!

src/main/resources/META-INF/resources/webjars/form-flow/0.0.1/css/libraryStyles.css
```diff
@@ -279,6 +279,18 @@ url('/webjars/form-flow/0.0.1/fonts/MaterialIcons-Regular.svg') format('svg');
...
+  content: '';
+  display: block;
+  position: absolute;
+  left: 0;
+  top: .25em;
+  width: 2.5rem;
+  height: 1.5rem;
+  background-image: url('/webjars/form-flow/0.0.1/images/emojis/checkmark.png');
+  background-repeat: no-repeat;
+  background-size: auto 1.5rem; }
```
coltborg, line 292: Two things:
- `crossmark` should have it's own rule
- We only need to override the one property, all others will still be applied from `honeycrisp.min.css`

```suggestion
.list--checkmark li:before {
  background-image: url('/webjars/form-flow/0.0.1/images/emojis/checkmark.png');
}

.list--crossmark li:before {
  background-image: url('/webjars/form-flow/0.0.1/images/emojis/crossmark.png');
}
```

coltborg (approved):
TY, lgtm :shipit:

PR #152, opened 2023-03-02
Add address verification screen live template

spokenbird:
[#184577011]


coltborg (approved):
Approved :shipit:. I tried all three live templates locally.

I made an adjustment to the state live template so that it showed in `.html` files.

PR #150, opened 2023-03-02
Adds onPost and crossValidationAction

bseeger:
Adds more actions to provide SNE ability to write custom code for pre validation as well as validation. 

https://www.pivotaltracker.com/story/show/184576570

PR #156, opened 2023-03-07
Adding "beforeDisplayAction"

sree-cfa:
This is copied from #150

Adding action hook to format data pulled from the DB before it is displayed to the client

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -91,6 +91,9 @@ public class ScreenController extends FormFlowController {
     if (currentScreen == null) {
       return new ModelAndView("redirect:/error");
     }
+
+    currentScreen.handleBeforeDisplayAction(submission, uuid);
+
     Map<String, Object> model = createModel(flow, screen, httpSession, submission);
     String formAction = createFormActionString(flow, screen);
     model.put("formAction", formAction);
@@ -246,6 +249,8 @@ public class ScreenController extends FormFlowController {
       HttpSession httpSession
   ) {
     Submission submission = submissionRepositoryService.findOrCreate(httpSession);
+    var currentScreen = getScreenConfig(flow, screen);
+    currentScreen.handleBeforeDisplayAction(submission, uuid);
     Map<String, Object> model = createModel(flow, screen, httpSession, submission);
     model.put("formAction", String.format("/%s/%s/%s", flow, screen, uuid));
     return new ModelAndView(String.format("%s/%s", flow, screen), model);
@@ -415,6 +420,8 @@ public class ScreenController extends FormFlowController {
     Optional<Submission> submissionOptional = submissionRepositoryService.findById(id);
     if (submissionOptional.isPresent()) {
       Submission submission = submissionOptional.get();
+      var currentScreen = getScreenConfig(flow, screen);
+      currentScreen.handleBeforeDisplayAction(submission, uuid);
       model = createModel(flow, screen, httpSession, submission);
 
       var existingInputData = submission.getInputData();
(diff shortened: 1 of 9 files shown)


src/main/java/formflow/library/config/ScreenNavigationConfiguration.java
```diff
@@ -21,18 +21,57 @@ public class ScreenNavigationConfiguration {
   private String subflow;
   //TODO: Implement callback
   private String callback;
-  private String beforeSave;
```
sree-cfa, line 24: Doesn't _seem_ like we need to keep this string variable after we've setup the `Action`

bseeger: Yeah, I'd agree, but we are stuck with this because of how we load this class. We use Snake YAML and it overlays yaml onto the class.  Did this actually work with out it?   Maybe it just needs the setter.... hmmm...

src/main/java/formflow/library/config/ScreenNavigationConfiguration.java
```diff
@@ -21,18 +21,57 @@ public class ScreenNavigationConfiguration {
...
+    if (beforeDisplayAction != null) {
+      if (uuid == null || uuid.isEmpty()) {
+        beforeDisplayAction.run(submission);
+      } else {
+        beforeDisplayAction.run(submission, uuid);
+      }
+    }
+  }
+
+  public void handleBeforeDisplayAction(FormSubmission formSubmission) {
```
bseeger, line 72: I think this parameter also needs to be Submission, not FormSubmission.

src/main/java/formflow/library/config/submission/Action.java
```diff
@@ -25,4 +28,23 @@ public default void run(Submission submission) {
...
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
+  /**
+   * Runs an action on form submission data to potentially manipulate the data.
+   *
+   * @param formSubmission form submission object the action is associated with, not null
+   * @param id             id for the iteration
+   */
+  public default void run(FormSubmission formSubmission, String id) {
```
bseeger, line 47: The more I think about this, the more I realize that the FormSubmission actions probably don't need the id part (for uuid) for running an action. But I could be wrong here. Maybe it's better to be flexible and provide this option.  Thoughts?

src/main/java/formflow/library/ScreenController.java
```diff
@@ -91,6 +91,9 @@ ModelAndView getScreen(
     if (currentScreen == null) {
       return new ModelAndView("redirect:/error");
     }
+
+    currentScreen.handleBeforeDisplayAction(submission, uuid);
```
bseeger, line 95: I _think_ we need this in the Post as well, maybe after line 160?  Not sure. 

The only reason (that I can think of) that we'd need this is if we display data from previous pages, we will want to make sure they are presentable.  Thoughts?

src/test/java/formflow/library/framework/BeforeDisplayTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @Test
+  void shouldSaveFormattedDate() throws Exception {
+    postExpectingSuccess("inputs",
+        Map.of("ssnInput", List.of("111-00-1234")));
+    assertThat(submission.getInputData().get("ssnInputEncrypted")).isEqualTo("BBB-AA-BCDE");
+    Assertions.assertNull(submission.getInputData().get("ssnInput"));
+
+    getPageExpectingSuccess("inputs");
+    assertThat(submission.getInputData().get("ssnInput")).isEqualTo("111-00-1234");
+  }
```
bseeger, line 56: It would be worth testing here to make sure the page's data doesn't have "ssnInputEncrypted" and that submission data on line 51 doesn't have "ssnInput".

src/main/java/formflow/library/config/ScreenNavigationConfiguration.java
sree-cfa: Yup :) - Works without it and just needs the setter

src/main/java/formflow/library/config/submission/Action.java
sree-cfa: Yeah - Flexibility would nice but I dont think we'll need it for the encryption workaround so I'll leave it out for now

src/test/java/formflow/library/framework/BeforeDisplayTest.java
sree-cfa: Er- how do I get the page's data here?

src/main/java/formflow/library/config/ScreenNavigationConfiguration.java
bseeger: Nice - I'm taking these out of the other PR as well.

bseeger (approved):
Looks good! Thanks for the changes

PR #157, opened 2023-03-08
Add method to check if entered address matches validated address

spokenbird:
[#184618630]


src/main/java/formflow/library/inputs/AddressParts.java
```diff
@@ -1,10 +1,19 @@
...
-  public static final String STREET_ADDRESS_2 = "StreetAddress2";
-  public static final String CITY = "City";
-  public static final String STATE = "State";
-  public static final String ZIPCODE = "ZipCode";
+  STREET_ADDRESS_1("StreetAddress1"),
+  STREET_ADDRESS_2("StreetAddress2"),
+  CITY("City"),
+  STATE("State"),
+  ZIPCODE("ZipCode");
+  private final String value;
```
bseeger, line 9: I like how you revamped this - it is much more inline with what an enum class should be. Instead of `value`, `part` might be more appropriate, but I don't feel strongly about this.

PR #158, opened 2023-03-09
Adding a test scenario for using a property in an Action

sree-cfa:
See related https://github.com/codeforamerica/la-doc-uploader/pull/40


bseeger:
HI @sree-cfa, is this still a work in progress?

bseeger:
Thanks, @sree-cfa - I don't think we will use this at this point.

PR #161, opened 2023-03-10
Move address validation to after field validation

coltborg:
[#184672973](https://www.pivotaltracker.com/story/show/184672973)


bencalegari (approved):

bencalegari:
@coltborg After I approved it just now I realized we should really have a test for this to enforce that order

bencalegari (approved):

PR #164, opened 2023-03-14
Change submission_id to be a UUID

spokenbird:
- Switch from H2 to Postgres in test
- Remove distinction in uuid generation since everything uses psql

#184056302


build.gradle
```diff
@@ -88,7 +88,8 @@ dependencies {
     testImplementation 'io.github.bonigarcia:webdrivermanager:5.3.2'
     testImplementation 'org.awaitility:awaitility'
     testImplementation 'org.seleniumhq.selenium:selenium-java'
-    testImplementation 'com.h2database:h2'
+//    testImplementation 'com.h2database:h2'
+    testImplementation 'org.postgresql:postgresql'
```
bseeger, line 92: Why are we pulling H2 and replacing it with postgresql?  The beauty of H2 to is that it is an in memory database that doesn't affect the engineers db.

spokenbird: We couldn't get H2 to play nice with the SQL needed for this migration. 

@bencalegari might be able to speak with more specifics on that.

bseeger: H2 will definitely make us jump through some hoops, sql statement wise, but with out H2 we will now need to make sure anyone running this test code has a db setup independently from our code - like github.

bseeger: @bencalegari which statements in Sql are having trouble being translated to H2 Sql?

bencalegari:
```sql
update user_files uf set new_submission_id = s.new_id from submissions s where s.id = uf.submission_id;
```
^ This fails on H2 saying `NULL not allowed for column s.id`. This is weird for many reasons, the biggest of which is that we aren't updating the submission id. Could just be a bunk error message but whether we used a placeholder for the UUID  generation function or called the H2-specific one, we got this result. H2's postgres compatibility seems insufficiently compatible. The `@Transactional` and the  `spring.jpa.hibernate.ddl-auto:create-drop` setting clean the database on test runs, leaving only the migrations. We should document the necessity to create a test db in this pr if we continue down this road. 

I really didn't want to do switch from H2 but I'm not aware of another way to phrase that necessary update in SQL. One alternative that just came to mind we didn't try was to write a [Java-based migration](https://flywaydb.org/documentation/tutorials/java#:~:text=Creating%20a%20Java%2Dbased%20migrations) but depending on how much we have to rely on underlying SQL statements instead of JDBC helpers that may not buy us enough to avoid this problem.

bseeger: Random thought:  We could also just skip that line in H2, as we have no data to actually migrate in the test scenario.

bseeger:
I saw a presentation this weekend that mentioned [`testContainer](https://www.testcontainers.org/). Might be worth checking this out, because then we could use postgres more ephemerally.

src/main/resources/application-form-flow-library.yaml
```diff
@@ -1,4 +0,0 @@
-spring:
-  flyway:
-    placeholders:
-      uuid_function: "gen_random_uuid"
```
bseeger, line 4: Taking this out means that V4 migration will not work. I think we need to update the V4 migration as well (just to remove the variable), I wouldn't change anything else.

bencalegari:
Change submission_id to be a UUID

- Tests now use postgresql

[#184056302]

Co-authored-by: Alex Gonzalez <agonzalez@codeforamerica.org>
Co-authored-by: Bethany Seeger <bseeger@codeforamerica.org>
Co-authored-by: Chibuisi Enyia <cenyia@codeforamerica.org>

src/main/resources/db/migration/V5__submission_id_change_primary_key_to_uuid.sql
```diff
@@ -0,0 +1,17 @@
...
+update user_files uf set new_submission_id = s.new_id from submissions s where s.id = uf.submission_id;
+alter table user_files drop constraint user_files_submission_id_fkey;
+
+alter table submissions drop column id;
+alter table submissions rename column new_id to id;
+
+alter table user_files drop column submission_id;
+alter table user_files rename column new_submission_id to submission_id;
+alter table user_files add constraint user_files_submission_id_fkey FOREIGN KEY (submission_id) REFERENCES submissions (id);
+
```
bseeger, line 15: I think we should create the test db in this migration.  That makes it easier for the user.

PR #170, opened 2023-03-24
Moving default encoding to application properties

sree-cfa:
Setting the default encoding for message resources in the application yaml instead of in the code so that it can be overridden as needed.

Verified this works with [@ladocuploader](https://github.com/codeforamerica/la-doc-uploader) on Transifex PR --> https://github.com/codeforamerica/la-doc-uploader/pull/60

**References**
[Spring common application properties](https://docs.spring.io/spring-boot/docs/1.4.1.RELEASE/reference/html/common-application-properties.html)


src/main/resources/application.yaml
```diff
@@ -1,4 +1,6 @@
 spring:
+  messages:
+    encoding: UTF-8
```
sree-cfa, line 3: Doesn't seem to affect downstream apps - they will use ISO as default for .properties files

src/main/resources/application-form-flow-library-test.yaml
```diff
@@ -3,3 +3,5 @@ spring:
     baselineOnMigrate: true
     placeholders:
       uuid_function: "random_uuid"
+  messages:
+    basename: messages,messages-form-flow
```
coltborg, line 7: Nitpick, add a space after the comma.
```suggestion
    basename: messages, messages-form-flow
```

coltborg (approved):
Lgtm; Ran it locally and it encoded correctly for UBI starter app. 

Added a nick pick comment, feel free to fix or ignore.

PR #176, opened 2023-04-05
Add setup script

coltborg:


setup.sh
```diff
@@ -0,0 +1,63 @@
+set -e
```
coltborg, line 1: @bencalegari I ran this and everything worked successfully! `brew update` took a long time, but I haven't ran it in awhile and that's somewhat expected.

Do you think we'd have another setup script in the starter app or would we point to this script in the starter app?

setup.sh
```diff
@@ -0,0 +1,63 @@
...
+log_line '--- FormFlow Setup Script ---'
+
+if ! brew --version; then
+  curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh
+else
+  brew update
+fi
+
+log_line 'Installing brew packages'
+brew install openjdk@17 jenv gradle postgresql@14
```
coltborg, line 24: Do we need to install `openjdk@17` is it playing the same role as `temurin-17` in `jenv`? Or is it needed for something else?

setup.sh
bencalegari: The latter seems sufficient for now. Glad it worked!

setup.sh
bencalegari: Good catch, that's a mistake I accidentally carried over to the script.. I can just choose Temurin and change the package.

bencalegari (approved):

PR #178, opened 2023-04-10
Update cross field validation tests to match contact info page scenario

spokenbird:
[#184673912]


src/test/java/formflow/library/inputs/TestFlow.java
```diff
@@ -53,14 +53,14 @@ public class TestFlow extends FlowInputs {
...
+  @Email(message = "Please enter a valid email address.")
+  String email;
 
   String phoneNumber;
 
-  String contactMethod;
+  ArrayList<String> howToContactYou;
 
-//    Old approach
+  //    Old approach
```
coltborg, line 63: These comments from line 63-71 can be deleted

src/test/java/formflow/library/submissions/actions/CheckIndicatedContactMethodIsProvided.java
```diff
@@ -2,27 +2,34 @@
 
 import formflow.library.config.submission.Action;
 import formflow.library.data.FormSubmission;
-
+import java.util.ArrayList;
```
coltborg, line 5: Why is there a `CheckIndicatedContactMethodIsProvided` in both the library and [the starter app](https://github.com/codeforamerica/form-flow-starter-app/blob/fd10854d2e1cccdc56a05042848f3f8f211f6b9f/src/main/java/org/formflowstartertemplate/app/submission/actions/CheckIndicatedContactMethodIsProvided.java)?

Is it for testing? Would the starter app use this one in the library?

coltborg:
Do you think other apps (doc-la, home-ebt) will need to make any updates once this PR is merged and deployed to maven?

src/test/java/formflow/library/submissions/actions/CheckIndicatedContactMethodIsProvided.java
spokenbird: This one is for testing the implementation of a CrossFieldValidation action. It matches the starter-app because that was the use case this was originally needed for so I just made the test for that specific use case. The thought was that the library should have a test for each style of action we have.

coltborg: Gotcha, makes sense.

coltborg (approved):
lgtm, tested locally

PR #180, opened 2023-04-11
Remove disabled and selected from selectOptionPlaceholder

coltborg:
If `selected` and `disabled` is used in `selectOptionPlaceholder`, it will not allow an empty string value to be passed to the server.

If only `selected` is used, there will be multiple `selected` for when refreshing/going back.

By removing both and keeping `selectOptionPlaceholder` as the first option, it will pass that value if a client does not change to a different option.


spokenbird (approved):
Tested and works in the starter app. Looks good to go! :)

PR #194, opened 2023-04-25
Format dates from DB fields in DataBaseFieldPreparer

spokenbird:
[#185026196]

### DIFF
--- a/src/main/java/formflow/library/pdf/DataBaseFieldPreparer.java
+++ b/src/main/java/formflow/library/pdf/DataBaseFieldPreparer.java
@@ -1,8 +1,12 @@
 package formflow.library.pdf;
 
 import formflow.library.data.Submission;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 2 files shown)


src/main/java/formflow/library/pdf/DataBaseFieldPreparer.java
```diff
@@ -24,14 +28,20 @@ public List<SubmissionField> prepareSubmissionFields(Submission submission) {
...
+        case "updatedAt" -> databaseFields.add(new DatabaseField(fieldName, formatDateTime(submission.getUpdatedAt())));
         case "flow" -> databaseFields.add(new DatabaseField(fieldName, submission.getFlow()));
         default -> log.error("Unable to map unknown database field: {}", fieldName);
       }
     });
     return databaseFields;
   }
+
+  private String formatDateTime(Date date) {
+    DateTimeFormatter formatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.SHORT).withLocale(Locale.US).withZone(
```
bseeger, line 43: This all looks good. One question this raises for me is the locale part. We might need to configure the system for a default timezone to work in that's separate from the timezone the system is running in.

bseeger: Could you please make the year 4 digits?  The DateTimeFormatter allows for patterns like this:  
`DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MM/dd/yyyy");`

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/format/DateTimeFormatter.html

PR #195, opened 2023-04-25
Add `currentSubflowItem` to get attributes of the current item

tdooner:
This allows pages within a subflow to get attributes that were previously submitted onto the current subflow item.


tdooner:
Probably should write some tests for this????

bseeger:
Yeah, that's a good idea.   This PR works great - I rebased it when testing it.

tdooner:
@bseeger I'd welcome some thoughts on the best way to test this since I didn't see anything that appeared to be testing the controller/view interface in ScreenControllerTest. Should it be in there?

tdooner:
@bseeger I took a stab at a test. Take a look and let me know what you think.

PR #196, opened 2023-04-25
Add `noneOfTheAbove` boolean to `checkboxInSet` input

tdooner:
In order to create a checkbox list where the last item is "None of the Above", this commit allows users of the library to pass `noneOfTheAbove=true` for the last checkbox. (Note that the honeycrisp JS will also need to be enabled with `noneOfTheAbove.init()`.)


bseeger (commented):
This looks good.  Is it worth documenting the honeycrisp.init() part in the readme?

tdooner:
@bseeger Done

PR #204, opened 2023-05-08
Call email service 184899355

coltborg:


src/main/java/formflow/library/email/MailgunEmailClient.java
```diff
@@ -0,0 +1,131 @@
...
+        Message builtMessage = message.build();
+
+        try {
+            mailgunMessagesApi.sendMessage(mailgunDomain, builtMessage);
+        } catch (FeignException exception) {
+            log.error(exception.getMessage());
+        }
+    }
+
+    public void setMailgunMessagesApi(MailgunMessagesApi mailgunMessageApi) {
```
bseeger, line 128: Could you please add a javadoc to this?

PR #207, opened 2023-05-08
Add X-forwarded-for headers filter

spokenbird:
[#184553834]


coltborg (approved):
Looks to be the same approach as HomeEBT. Looks good.

PR #208, opened 2023-05-10
Support "go back" in a subflow iteration

bseeger:
Helps to address: https://www.pivotaltracker.com/story/show/183082612

Revamps the code so that going back a page in a browser works for subflows that are greater that one page. 

This solidifies the use of "currentSubflowItem" in the Model to represent the current subflow's data.  `inputData`, in the Model, sometimes represented iteration data and sometimes the full input data.  Now it will always be all the input data and doesn't change meaning.


coltborg (approved):
Approved, I pulled it locally and did another pass for the whole flow!

PR #212, opened 2023-05-11
Delete tmp PDF files (was: Create TempFileDeletionScheduler)

spokenbird:
addresses: https://www.pivotaltracker.com/story/show/185152963

Instead of making a schedule to delete the files, I opted to delete them just before being shared back to the caller.


src/main/java/formflow/library/PdfController.java
```diff
@@ -24,31 +24,43 @@
...
-            PdfFile filledPdf = pdfGenerator.generate(flow, UUID.fromString(submissionId));
-            filledPdf.finalizeForSending();
-            HttpHeaders headers = new HttpHeaders();
-            headers.add(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=%s-%s.pdf".formatted(filledPdf.name(), submissionId));
-            return ResponseEntity.ok().contentType(MediaType.APPLICATION_OCTET_STREAM).headers(headers).body(filledPdf.fileBytes());
-        } else {
-            log.error("Attempted to download PDF with submission_id: " + submissionId + " but session_id was: " + httpSession.getAttribute("id"));
-            return ResponseEntity.status(HttpStatus.FORBIDDEN).body(messageSource.getMessage("error.forbidden", null, null));
-        }
+      headers.add(HttpHeaders.CONTENT_DISPOSITION,
```
bseeger, line 51: It looks like I changed a lot, but the only real changes are this spot here, where we grab the byte buffer, delete the file and then return the buffer. There is another spot I will mark.

src/main/java/formflow/library/pdf/PdfFile.java
```diff
@@ -14,37 +15,44 @@
...
+  }
+
+  public void finalizeForSending() throws IOException {
+    PDDocument pdDocument = PDDocument.load(fileBytes());
+    pdDocument.getDocumentCatalog().getAcroForm().flatten();
+    pdDocument.save(path);
+    pdDocument.close();
+  }
+
+  public void deleteFile() throws IOException {
```
bseeger, line 55: I added this method.

src/main/java/formflow/library/pdf/PdfFile.java
```diff
@@ -14,37 +15,44 @@
...
-        pdDocument.save(path);
-        pdDocument.close();
+  // put all the file in the same tmp java location
+  private static final File tmpFileDir = new ApplicationTemp().getDir();
+
+  public static PdfFile copyToTempFile(String pathToResource) {
+    InputStream unfilledPdf = PdfFile.class.getResourceAsStream(pathToResource);
+    Path resourcePath = Path.of(pathToResource);
+    String fileNameWithExtension = resourcePath.getFileName().toString();
+    String fileExtension = fileNameWithExtension.substring(fileNameWithExtension.lastIndexOf('.'));
```
bseeger, line 25: I didn't want to make assumptions about file extension, so I moved a few things around here.  Also just made it take the fileName and fileExtension as parameters to the createFileTemp. It makes for a cleaner temp file: 

`Multipage-UBI-Form3307367797630564204.pdf`

Versus: 

`testPdf.pdf17812849852771651595.pdf`

src/test/java/formflow/library/pdf/PdfFileTest.java
```diff
@@ -14,33 +15,93 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 class PdfFileTest {
-    PdfFile tempPdfFile;
```
bseeger, line 17: @spokenbird we seem to have different formatting rules - let's look into this when we have a chance.

bencalegari (approved):

PR #216, opened 2023-05-17
Updating logging and refactor

sree-cfa:
Updating logging error message to be a little more detailed. 
Refactoring to consolidate error message


bseeger (approved):
Looks good to me. Thanks, @sree-cfa

PR #218, opened 2023-05-18
catch array out of bounds error and throw informative error message i…

lkemperman-cfa:
…n case of flow name not existing


src/main/java/formflow/library/ScreenController.java
```diff
@@ -15,13 +15,9 @@
 import formflow.library.inputs.UnvalidatedField;
 import jakarta.servlet.http.HttpSession;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
bseeger, line 18: Hi @lkemperman-cfa could you reverse this?   We have a preference of not using ".*" for includes. Though I can see we are using many things from java.util... so I might need to rethink this.  I think being specific keeps the size down for what's included, but I need to double check that assumption.

lkemperman-cfa: Whoops, I'm sorry I think my IDE did that :D did not intend to make that change, thanks for pointing out

bencalegari (approved):

PR #228, opened 2023-05-31
Generate pdf service class

enyia21:


src/main/java/formflow/library/PdfService.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Uses pdfGenerator to generate a pdf for a flow using a UUID.
+   *
+   * @param flow
+   * @param submissionId
+   * @return a pdf byte array
+   */
+  public byte[] getFilledOutPDF(String flow, String submissionId) throws IOException {
+    // 1. generate the pdf
```
bseeger, line 34: Can you take out this above comment?

PR #231, opened 2023-06-01
Error messages [#185195335]

bencalegari:


bseeger:
This seems to work in some cases but lose context of the error in others: 

`http://localhost:8080/flow` - got a 404, good. 
`http://localhost:8080/flow/ubi` - got a 404, good. 
Screenshot 2023-06-02 at 11 30 55 AM

`http://localhost:8080/flow/ubi/nopage` - got a 999 and the URL changed to `http://localhost:8080/error`


The third scenario is the most likely scenario will run into a 404 with as developers work with the system.

src/main/java/formflow/library/controllers/FormFlowErrorController.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private final boolean showStackTrace;
+  private final String errorPath;
+
+  private final String prettyPrintPackages;
+  private final ErrorAttributes errorAttributes;
+
+  public FormFlowErrorController(
+      @Value("${form-flow.error.show-stack-trace:false}") boolean showStackTrace,
+      @Value("${server.error.path:${error.path:/error}}") String errorPath,
```
bseeger, line 32: Is this default of `${error.path:/error}` - where is `error.path` to be defined? Is it a system thing?

README.md
```diff
@@ -1586,6 +1586,13 @@ form-flow:
   path: 'name-of-file.yaml'
 ```
 
+#### Error properties
+
+| Property                                | Value                    | Default    | Description                                                                                                        |
+|-----------------------------------------|--------------------------|------------|--------------------------------------------------------------------------------------------------------------------|
+| `form-flow.error.show-full-error`       | `boolean`                | `true`     | Show full the full stack trace and error codes on our '/error' template (on your local or staging env for example) |
```
coltborg, line 1593: I think `show-full-error` here is meant to be `show-stack-trace`?

src/main/java/formflow/library/controllers/FormFlowErrorController.java
```diff
@@ -0,0 +1,88 @@
+package formflow.library.controllers;
```
bseeger, line 1: I was curious as to why you put this controller in a `controllers` directory when the rest of our controllers are not in that directory? Maybe we should move them all?

src/main/java/formflow/library/controllers/FormFlowErrorController.java
```diff
@@ -0,0 +1,88 @@
...
+  @RequestMapping("/error")
+  public ModelAndView handleError(WebRequest webRequest) {
+    String prettyException = prettyPrintStackTrace(errorAttributes.getError(webRequest));
+
+    Map<String, Object> errorAttributesMap = errorAttributes.getErrorAttributes(
+        webRequest,
+        ErrorAttributeOptions.defaults().including(BINDING_ERRORS, EXCEPTION, MESSAGE, STACK_TRACE)
+    );
+    errorAttributesMap.put("prettyStackTrace", prettyException);
+    if (showStackTrace) {
```
coltborg, line 50: When I create a thymeleaf error and turn `showStackTrace` off, I assume it shows the apache error screen? Which does include a stack trace. With the name, I was assuming there wouldn't be a stack trace on the page.

Maybe that's just a name expectation I'm having? Should the variable name be updated or is this the expected behavior?

Screenshot 2023-06-02 at 10 44 23 AM

src/main/java/formflow/library/ScreenController.java
```diff
@@ -237,6 +234,11 @@ ModelAndView getSubflowScreen(
     return new ModelAndView(String.format("%s/%s", flow, screen), model);
   }
 
+  @NotNull
+  private static ModelAndView errorScreen() {
+    return new ModelAndView(new RedirectView("/error"));
```
bseeger, line 239: Are we able to pass more context into this, like the "BAD_REQUEST" info? It seems like we are losing that information when generating the error page.

bencalegari: I can add it back in but I wasn't sure what we we're getting by reporting it. If we have the the full trace on we'd see all the relevant information, if not we'd just get redirected to the client facing error page. Am I missing something?

src/main/java/formflow/library/controllers/FormFlowErrorController.java
bencalegari: This one's bad copy paste, thanks for catching that!

src/main/java/formflow/library/controllers/FormFlowErrorController.java
bencalegari: I noticed this too earlier on in my journey through this ticket but when I tried that (and updated the path reference to `ScreenController` in `build.gradle`) it didn't work so I abandoned it but forgot to move this out. I should see if I can get it working with all but the `ScreenController` and if not I'll put this at the top level as well.

src/main/java/formflow/library/ScreenController.java
bseeger: In the case that the user went to a page in the flow that didn't exist `/flow/ubi/junkPage`, the error page didn't report anything useful and had a 999 error (image above).   There was no stack trace either.  Mostly, it'd be good to have better error reporting in that scenario. So, not sure if passing in more context would work, but definitely having more messaging in that one scenario would be good.

src/main/java/formflow/library/controllers/FormFlowErrorController.java
coltborg: @bencalegari what was this suppose to be? Right now it's only hitting the default (which is working fine). Do you mean `server.error.path` should be something like `form-flow.error.path`?

src/main/java/formflow/library/controllers/FormFlowErrorController.java
coltborg: Moved this to out of `controllers` directory.

src/main/java/formflow/library/controllers/FormFlowErrorController.java
bencalegari: I wanted `@Value("${server.error.path:/error")`, so it can be overridden to a different error path if you wanted to implement your own.

coltborg: Updated!

src/main/java/formflow/library/ScreenController.java
coltborg: So the approach of using `redirect:` was not allowing status code or anything else to get passed to the `/error` controller mapping.

I started a [WIP approach of creating a model of info](https://github.com/codeforamerica/form-flow/pull/231/files#diff-e3c0b2bc250fdece31f72615351249643e60c4ce9dff6cad2730fb9ea87dc9efR240-R251) needed to pass to the error page but this approach essentially repeats much of the same code that lives in `FormFlowErrorController` 😢 

Open to alternative ideas, hence why I didn't try and finish all of the model creation.

coltborg: Here are dev notes of the things I tried to resolve this issue today, just for reference.

---

* `errorScreen()` private method in the ScreenController is not being called for `/asdf` or `/flow/asdf`. Only for `/flow/ubi/asdf`
* I don't thing `/error` is being called with a `redirect:/error` in the first two that are working
* When we add the status code to ModelAndView, it doesn't even get to `/error`!
- [x] Does using `forward:/error` working?
	* No, it does the same thing as `redirect:`
- [x] Is there a code in the `WebRequest`?
	* No
- [x] What if I add an exception handler to the screen controller?
	* Couldn't get it to work
- [x] There's only `WebRequest` in the error controller, what if there's something I can get from `httpServletRequest`?
	* No, I still can't get the passed in status
- [x] Did the Readme change happen?
	* Yep, needed to change it
- [x] Move file out of `controller` folder since `ScreenController` doesn't work from there
- [x] I should move `error` and `devError` into their own folder in `resources/templates`
- So using `redirect:/error` doesn't work and is the cause of swallowing up the `404` status
- I either need a better way to to move from the `ScreenController` to the error controller endpoint or I need to do the same work and populate `BINDING_ERRORS, EXCEPTION, MESSAGE, STACKTRACE`  in `errorScreen()`

---

coltborg: Now by throwing an exception, we can redirect to our `/error` endpoint with the right status codes being sent!

src/main/java/formflow/library/ScreenController.java
```diff
@@ -434,20 +436,32 @@ ModelAndView deleteSubflowIteration(
...
+  private static void errorScreen(String method, String url, HttpStatus httpStatus) throws Exception {
+    HttpHeaders headers = new HttpHeaders();
+
+    if (httpStatus.isSameCodeAs(HttpStatus.NOT_FOUND)) {
+      throw new NoHandlerFoundException(method, url, headers);
+    }
+
+    if (httpStatus.isSameCodeAs(HttpStatus.BAD_REQUEST)) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST);
+    }
```
bseeger, line 464: This code update is great!  My one thought is should we have it thrown a generic exception if it's not NOT_FOUND or BAD_REQUEST?  Not sure what it would be, but it seems like we should have some sort of generic reporting in that scenario so we don't have a silent failure.

coltborg: Ya, I'll add a default exception that passes what it's given

coltborg: Actually, `ResponseStatusException` is a good default. Since that's only a one liner, I refactored the places that want to throw a `BAD_REQUEST`, to throw in place. Then I refactored `errorScreen` to `throwNotFoundError` and it doesn't have control logic.

PR #237, opened 2023-06-06
Encryption service

sree-cfa:
Adding an encryption service for encrypting sensitive data

PT#185334981


src/main/java/formflow/library/data/SubmissionEncryptionService.java
```diff
@@ -0,0 +1,115 @@
...
+  public Submission encrypt(Submission submission) {
+    if (submission.getFlow() == null) {
+      return submission;
+    }
+
+    try {
+      Class<?> flowClass = Class.forName(inputConfigPath + StringUtils.capitalize(submission.getFlow()));
+      Map<String, Object> inputData = submission.getInputData();
+
+      List<Field> fields = Arrays.stream(flowClass.getDeclaredFields()).filter(field -> Arrays.stream(field.getAnnotations()).anyMatch(annotation -> annotation.annotationType().getName().endsWith("Encrypted"))).toList();
```
bseeger, line 52: Is the annotation name a complete class path and so we only care what it ends with?  I wonder if we want (dot) Encrypted (`.Encrypted`) instead, so we can be sure it's the "Encrypted" class and not "SomeOtherMeaningEncrypted"?

PR #241, opened 2023-06-08
Add PDF Generation Documentation

spokenbird:


README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+is `Yes`.
+
+### Creating a pdf-map.yaml file
+
+The next step in generating PDFs is to create a `pdf-map.yaml` file. This file will act as a map of
+input fields from your application to PDF fields in your template PDF file. This file should be
+added
+to your applications `resources` folder, in a folder named `pdfs`.
+
+Your `pdf-map.yaml` file will consist of the following key value pairs:
```
bseeger, line 629: Instead of leaving this as a `:`, it might be better to reword this, like so:
`The following sections describe the expected keys in the `pdf-map.yaml` file.`

README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+    incomeUnemployment: INCOME_HAS_UNEMPLOYMENT_MEMBER
+```
+
+Where `incomeTypes` is the name of a checkbox input in your application,
+and `incomeJob`, `incomeSelf`,
+and `incomeUnemployment` are the actual values of checkboxes within that
+input. `INCOME_HAS_JOB_MEMBER`,
+`INCOME_HAS_SELF_EMPLOYMENT_MEMBER`, and `INCOME_HAS_UNEMPLOYMENT_MEMBER` are the names of the
+corresponding PDF checkbox fields.
+
```
bseeger, line 769: I think it might be useful to show an example of a while file here - the culmination of all these sections of details.  And maybe a link to the file in the starter app.  Seeing all this in context will definitely help people understand things better.

README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+### Custom preparers
```
bseeger, line 770: I think it's worthwhile and necessary to talk about our default prepares so folks know what they get for "free" with our work. It will also help explain what a custom preparer does.  `The custom prepare will allow for overriding of the default preparers work as well as allows for injecting new fields, etc...`

In other words, understanding what we do for them to start with will help explain why custom preparers are useful.  They then will know that they don't have to do any default preparing, just any custom work they want.

bseeger: I know you touch on this a little, but I think it's worth describing them more before you talk about the custom ones.

README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+
+When creating a custom preparer, you will want to create a new Java class that implements the
+`SubmissionFieldPreparer` interface and it's `prepareSubmissionFields` method.
+
+Depending on what you would like to do with your custom preparer, it should return a map of String
+input name to either `SingleField`, `CheckboxField`, `DatabaseField` or even a combination.
+
+For example, if you wanted to create a custom preparer that takes three seperate date fields and
+maps
+them to a single PDF
```
bseeger, line 803: "them to a single PDF field"

README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+Note that you would need to create a field for this input in your `pdf-map.yaml` file. For example:
+
+```yaml
+inputFields:
+  applicantDateOfBirth: APPLICANT_DATE_OF_BIRTH
+```
+
+### Custom Preparers for Subflows
+
+Working with subflows is a bit different for custom preparers. In subflows, you will need to append
```
bseeger, line 850: I think we need to go into the "why" a little bit, explaining that we need to flatten the iteration data into something the PDF understands. Right now it's a list of maps in the inputData, but we need to map all that into the PDF field mapping, which has to be really specific as to where those fields map into. 

This should also talk about how we are expecting the PDF template file mappings to use the `_n` paradigm as well. We are making an assumption that the PDF file is using this pattern and we need to be clear to folks about that. 

Maybe some of that is best done in the  "Creating a template PDF file" section

README.md
```diff
@@ -579,6 +580,314 @@ Note that the subflows are an array of repeating entries - one for each iteratio
...
+
+```yaml
+subflowInfo:
+  household:
+    totalIterations: 5
+```
+
+#### fields
+
+Similar to `inputFields`, `fields` is a map of input fields from your application to PDF fields in
```
coltborg, line 736: So `fields` is the same thing as `inputFields`, just in the `subflowInfo` key? Could it be `inputFields`? Since it would be `subflowInfo.<subflowName>.inputFields`?

It's complicated either way, but my first thought was "what the heck are fields"?

PR #242, opened 2023-06-08
Remove data action and subflow list from subflow preparer

spokenbird:


bseeger:
This looks good so far, but you should probably take out the action as well, since we won't use it. 

- https://github.com/codeforamerica/form-flow/blob/5ce01e95fbf7c289889e35851bbe662c400996aa/src/main/java/formflow/library/config/submission/Action.java#L60
- https://github.com/codeforamerica/form-flow/blob/5ce01e95fbf7c289889e35851bbe662c400996aa/src/test/java/formflow/library/pdf/RemoveApplicantIterationAction.java#L19

PR #247, opened 2023-06-12
Encryption service key fix 🍖 

sree-cfa:
Adding support for not having an encryption key


bseeger:
Does this PR make it so that if you don't config this for encryption, the server will still start successfully?

sree-cfa:
> Does this PR make it so that if you don't config this for encryption, the server will still start successfully?

It has to be present but blank - but thats a good point. Let me make that change so it doesnt have to be defined at all

PR #256, opened 2023-06-14
Adds implementing class names to condition and action run error message

spokenbird:


src/main/java/formflow/library/config/submission/Action.java
```diff
@@ -72,15 +71,15 @@ public default void run(FormSubmission formSubmission, String id) {
    * @param submission     submission object the action is associated with, not null
    * @param id             id for the iteration
    */
-  public default void run(FormSubmission formSubmission, Submission submission, String id) {
+  default void run(FormSubmission formSubmission, Submission submission, String id) {
     throw new UnsupportedOperationException("Not implemented run Validation");
```
bseeger, line 75: Should this change to be more descriptive as well?

spokenbird: Added, thanks for catching that!

PR #261, opened 2023-06-16
Allow for address validation in subflows

spokenbird:


src/main/java/formflow/library/ScreenController.java
```diff
@@ -508,21 +509,19 @@ private Boolean isConditionalNavigation(ScreenNavigationConfiguration currentScr
...
    * @return List<NextScreen> list of next screens
    */
   private List<NextScreen> getConditionalNextScreen(ScreenNavigationConfiguration currentScreen,
-      HttpSession httpSession, String subflowUuid) {
+      Submission submission, String subflowUuid) {
     return currentScreen.getNextScreens().stream()
         .filter(nextScreen -> conditionManager.conditionExists(nextScreen.getCondition()))
         .filter(nextScreen -> {
           if (currentScreen.getSubflow() != null) {
-            return conditionManager.runCondition(nextScreen.getCondition(), submissionRepositoryService.findOrCreate(httpSession),
```
coltborg, line 521: Am I correct in thinking that the reason we are passing the `submission` into `getConditionalNextScreen` instead of `httpSession` because we were using `httpSession` to get the `submission` anyway in the method? So you're simplifying it in that way?

spokenbird: Yes and also because in some places we were getting it from the DB  before the most recent version had been saved to the DB so we opted to use the updated version instead of the DB version.

coltborg (approved):
lgtm

This is tough to test because there is no subflow that has address validation in a subflow 😵

PR #275, opened 2023-06-27
bumps gradle spring boot plugin version

bseeger:
Bumps from 3.0.5 -> 3.1.1.


bseeger:
This seems to only affect compilation of the library and has no affect on the version located in the starter apps file.  I'm bumping this to keep up with the version in the starter app.

PR #280, opened 2023-06-28
Create get in S3CloudFileRepository and UploadController.downloadFile

spokenbird:
[#185447786]


src/test/java/formflow/library/config/WebMvcConfig.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+@EnableAsync
+public class WebMvcConfig implements AsyncConfigurer {
+
+  @Override
+  public Executor getAsyncExecutor() {
```
bseeger, line 14: What purpose does this serve?  I'm nervous about something being async in testing, but not when running the standard code, so I'm not clear on what should be async in testing.

README.md
```diff
@@ -244,24 +247,37 @@ class Submission {
 
   @Id
   @GeneratedValue
-  private Long id;
+  private UUID id;
```
bseeger, line 250: Why is this showing up here? These submission changes have been in the code for months....????

README.md
```diff
@@ -1376,6 +1392,32 @@ bucket. This will automatically delete files in your bucket that are older than
...
+## Document Download
+
+Form-flow library allows users to either:
+
+1. Download individual files
+2. Download a zip folder with all the files associated with a submission.
+
+In order to download from any endpoint, the http session must have an attribute of "id" that matches
+the submission's UUID that you would like to download. There are
+
```
bseeger, line 1404: I'd change "any" to "these".  "http" should probably be "HTTP" since it's an abbreviation.  This just ends at "There are" ... should there be more?

README.md
```diff
@@ -1376,6 +1392,32 @@ bucket. This will automatically delete files in your bucket that are older than
...
+### Downloading individual files
+
+To downlodoad individual files a user must reach single-file download
+endpoint: `/file-download/{submissionId}/{fileId}`.  
+The `submissionId` must be included in the session as an attribute and the `fileId` must be found in
```
bseeger, line 1409: Instead of "reach" the word "use" might flow better.  "must" could just be "can".  "To download individual files a user can use the single-file download endpoint: ".

README.md
```diff
@@ -1376,6 +1392,32 @@ bucket. This will automatically delete files in your bucket that are older than
...
+the `user_files` table in order for the file to be retrieved.
+
+### Downloading all files
+
+In order to download all the files associated with a submission a user needs to use the download-all
+endpoint:  `/file-download/{submissionId}`. The download-all endpoint requires that the `id`
+attribute for the httpsession matches the `submissionId` for the group of files in the submission
+that
```
bseeger, line 1417: "httpsession" should probably be "HTTP session", for consistency's sake.

README.md
```diff
@@ -1376,6 +1392,32 @@ bucket. This will automatically delete files in your bucket that are older than
...
+you would like to download. When you reach the download-all endpoint every file in the `user_files`
+table associated with a submission are zipped together and downloaded.
```
bseeger, line 1419: "When you use", versus "reach" might flow better.   "reach" isn't a word I've heard applied to API's much.

src/main/java/formflow/library/FileController.java
```diff
@@ -50,6 +62,18 @@ public UploadController(
...
+  /**
+   * File upload endpoint.
+   *
+   * @param file         A MultipartFile file
+   * @param flow         The current flow name, not null
+   * @param inputName    The current inputName
+   * @param thumbDataUrl The thumbnail URL generated from the upload
+   * @param httpSession  The current httpSession
+   * @return ON SUCCESS: A ResponseEntity with a status of (HttpStatus.OK), a * contentType of (MediaType.TEXT_PLAIN) and a
+   * body(newFileId.toString()).
```
bseeger, line 74: For the `@return` - take out the "ON SUCCESS" (unless this is a java standard??).  Also, you can just say "ResponseEntity with a body containing the id of the file, if successfully uploaded, else an error code will be returned." here, no one will understand the "newFileId.toString()" -> that won't make sense to them unless they read the code, which they shouldn't have to. 

These javadocs should provide enough abstracted  information (word description, not code) about the function so that they don't have to read the code.

src/main/java/formflow/library/FileController.java
```diff
@@ -50,6 +62,18 @@ public UploadController(
     this.messageSource = messageSource;
   }
 
+  /**
+   * File upload endpoint.
+   *
+   * @param file         A MultipartFile file
+   * @param flow         The current flow name, not null
```
bseeger, line 69: I think all the params are required and should not be null, so maybe add that to all of them, or none?

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
       return new RedirectView("/error");
     }
   }
+
+  /**
+   * @param httpSession  - The current httpsession
```
bseeger, line 218: I don't think Javadocs like the hyphen (-) and they should probably be removed. 

Also, across this file, can we call be consistent with `httpsession`, in comments, and call it `HttpSession`, after the Java object, or more generally `HTTP Session` (generically what it is)?

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
...
+        .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + filename + "\"")
+        .contentType(MediaType.parseMediaType(contentType))
+        .contentLength(fileSize)
+        .body(responseBody);
+  }
+
+  /**
+   * @param httpSession  The current httpsession.
+   * @param submissionId The submissionId of the all the files that you would like to download.
+   * @return ON SUCCESS: A ResponseEntity with a status of ok, a header, and a responseBody that a zip folder with all the files
```
bseeger, line 275: "folder" should probable be "file" - it's a zip file. 

"assoicated" is misspelled.

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
       return new RedirectView("/error");
     }
   }
+
+  /**
+   * @param httpSession  - The current httpsession
+   * @param submissionId - The submissionId of the file to be downloaded
+   * @param fileId       - The UUID of the file to be downloaded.
+   * @return ON SUCCESS: A ResponseEntity with a status of ok, a header, the contentType, the content size, and a responseBody
+   * that includes the file.
```
bseeger, line 222: `responseBody` is the name of the variable, which requires someone to read the code to figure out what it actually us.  You could just call  say "response body".

src/main/java/formflow/library/FileController.java
bseeger: Here's an example of a rendered Javadoc and how these javadocs might look once we render them - this does not include the code, so referencing variables in the code is not helpful at all: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#replace(java.lang.CharSequence,java.lang.CharSequence)

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
...
+   * <p>ON FAILURE: A ResponseEntity with a status of forbidden-403 or not found- 404</p>
+   */
+  @GetMapping("/file-download/{submissionId}")
+  ResponseEntity<StreamingResponseBody> downloadAllFiles(
+      HttpSession httpSession,
+      @PathVariable String submissionId
+  ) {
+    Optional<Submission> maybeSubmission = submissionRepositoryService.findById(UUID.fromString(submissionId));
+    if (httpSession.getAttribute("id").toString().equals(submissionId) && maybeSubmission.isPresent()) {
+      if (maybeSubmission.isEmpty()) {
```
bseeger, line 286: Can it pass "isPresent()" but be empty??  This "if" should probably be the "else" to line 285.  If you get here, there is a submission, as you just checked "isPresent()"

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
...
+      StreamingResponseBody responseBody = outputStream -> {
+        baos.writeTo(outputStream);
+        baos.close();
+      };
+
+      return ResponseEntity.ok()
+          .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + "UserFiles-" + submission.getId() + ".zip" + "\"")
+          .body(responseBody);
+    } else {
+      log.error(
```
bseeger, line 323: Or the requested submission does not exist.... as per the maybeSubmission.isPresent() check being false. 

If the submission doesn't exist, it would be good to differentiate that in the error message we log.

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
...
+  ) {
+
+    Optional<UserFile> maybeFile = uploadedFileRepositoryService.findById(UUID.fromString(fileId));
+    if (maybeFile.isEmpty()) {
+      log.error(String.format("File with id %s may have already been deleted", fileId));
+      return ResponseEntity.notFound().build();
+    }
+
+    UserFile file = maybeFile.get();
+    if (!submissionId.equals(httpSession.getAttribute("id").toString())) {
```
bseeger, line 239: This would be better as the first thing you do (on line 232) - as you have the data handy and can short circuit any other work if this check fails.  Rather than do some look ups and then check data you didn't have to look up. 

Also would be good to log an error if we fail this check.

README.md
bseeger: I know this is innocuous, since we end up at the same result of where we are at now, but it is concerning to me that this PR looks like we are going from the changes on the left to the changes that are already on main.  Any clue how this happens?  

 https://github.com/codeforamerica/form-
flow/blob/main/src/main/java/formflow/library/data/Submission.java

There's something happening in someone's github flow that's resulting in a weird looking starting point and would be good to find out what that is, as that's where errors like regressions can creep in.

src/main/java/formflow/library/upload/S3CloudFileRepository.java
```diff
@@ -71,6 +75,25 @@ public void upload(String filePath, MultipartFile file) {
...
+
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      IOUtils.copy(inputStream, outputStream);
+      byte[] fileBytes = outputStream.toByteArray();
+      long fileSize = fileBytes.length;
+
+      log.info("File {} successfully downloaded", filepath);
+      return new CloudFile(fileSize, fileBytes);
+    } catch (IOException e) {
+      log.error("Exception occurred while attempting to get file in S3 code: " + e.getMessage());
```
bseeger, line 92: I think it's worthwhile to include more information here, unless the e.getMessage() will always include "filepath" in the event of an error? If not, including "filepath" in the log messages is good for debugging purposes.

src/main/java/formflow/library/FileController.java
```diff
@@ -181,4 +213,117 @@ RedirectView delete(
...
+  @GetMapping("/file-download/{submissionId}/{fileId}")
+  public ResponseEntity<StreamingResponseBody> downloadSingleFile(
+      HttpSession httpSession,
+      @PathVariable String submissionId,
+      @PathVariable String fileId
+  ) {
+
+    Optional<UserFile> maybeFile = uploadedFileRepositoryService.findById(UUID.fromString(fileId));
+    if (maybeFile.isEmpty()) {
+      log.error(String.format("File with id %s may have already been deleted", fileId));
```
bseeger, line 234: It will help us debug if this were to log more details  (submission id with the file id) and note that the user was trying to delete it, but we couldn't find the file.  Currently you are suggesting _what might have_ happened, but it would be easier to debug if we stated _what is happening_ ("Request to delete fileId (%d), but file not found in database.  (submission id %d)")

src/test/java/formflow/library/controllers/FileControllerTest.java
```diff
@@ -180,4 +190,113 @@ public void endpointDeletesFromCloudRepositoryDbAndSession() throws Exception {
...
+
+      byte[] testZipFile = Files.readAllBytes(Paths.get("src/test/resources/test-archive.zip"));
+
+      Map<String, byte[]> responseContents = getZipContentsMap(response);
+      Map<String, byte[]> testZipContents = getZipContentsMap(testZipFile);
+
+      assertThat(responseContents.keySet()).isEqualTo(testZipContents.keySet());
+
+      for (String name : responseContents.keySet()) {
+        assertThat(Arrays.equals(responseContents.get(name), testZipContents.get(name))).isTrue();
```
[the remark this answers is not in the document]
spokenbird: We tried to do this but found that the sizes aren't consistent between MockMVC's return and the original file. I think it's because the response includes the bytes from the response headers which add more bytes.

src/main/java/formflow/library/FileController.java
```diff
@@ -144,26 +172,26 @@ RedirectView delete(
...
       if (maybeFile.isEmpty()) {
         log.error(String.format("File with id %s may have already been deleted", fileId));
         return new RedirectView("/error");
       }
 
       UserFile file = maybeFile.get();
-      if (!submissionId.equals(file.getSubmission_id().getId())) {
-        log.error(String.format("Submission %d does not match file %s's submission id %d", submissionId, fileId,
-            file.getSubmission_id().getId()));
+      if (!submissionId.equals(file.getSubmissionId().getId())) {
```
bseeger, line 186: As Chibuisi mentioned, this is really confusing.  The naming is definitely misleading:`getSubmissionId` returning the `Submission`. Can we make this easier to work work with or are we stuck with this because of DB magic?  

Can we call variable `submissionId` `submission` instead, in the UserFile.java code?  That might work, since we specify the column name in an attribute - so it knows which column to grab. Perhaps the variable name in that file doesn't carry any semantic significance?

README.md
```diff
@@ -1376,6 +1392,32 @@ bucket. This will automatically delete files in your bucket that are older than
 permits.
 [You can read more about configuring a retention policy in S3 here.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html)
 
+## Document Download
+
+Form-flow library allows users to either:
```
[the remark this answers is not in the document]
spokenbird: We do say Form-flow in quite a few places actually, should I update it in all places?

src/main/java/formflow/library/data/UserFile.java
```diff
@@ -43,11 +43,11 @@ public class UserFile {
   @GeneratedValue
   //@Type(type = "org.hibernate.type.UUIDCharType")
   //@Type(type = "pg-uuid")
-  private UUID file_id;
+  private UUID fileId;
 
   @ManyToOne
   @JoinColumn(name = "submission_id")
-  private Submission submission_id;
+  private Submission submission;
```
bseeger, line 50: Whoa, did it work? If so, awesome!  Thanks for making that change, @spokenbird!

src/test/java/formflow/library/utilities/SeleniumFactory.java
```diff
@@ -36,7 +36,7 @@ public void start() {
     // Warning: you may need this line to get a later version of the chromedriver.
     // once you have the version, you don't seem to need to specify it anymore.
     // (or maybe you can upgrade your driver outside of this)
-    // WebDriverManager.chromedriver().driverVersion("111.0.5563.64").setup();
+    WebDriverManager.chromedriver().driverVersion("114.0.5735.198").setup();
```
bseeger, line 39: This should come out - I don't *think* we need it anymore.

PR #306, opened 2023-07-05
Adds journey test for address validation in a subflow

bseeger:
Helps satisfy: https://www.pivotaltracker.com/story/show/185152074


src/test/java/formflow/library/utilities/AbstractBasePageTest.java
```diff
@@ -46,6 +46,8 @@ public abstract class AbstractBasePageTest {
 
   @BeforeEach
   protected void setUp() throws IOException {
+    driver.manage().deleteAllCookies();
+    driver.resetInputState();
     initTestPage();
```
bseeger, line 51: I'm not 100% sure what the two lines above this are doing, but I wanted to clear out session information when `setup()` was called and this seems to do that.

src/test/resources/application-test.yaml
```diff
@@ -22,6 +22,8 @@ form-flow:
       domain: 'mail.forms-starter.cfa-platforms.org'
       sender-email: 'Testing <test@mail.forms-starter.cfa-platforms.org>'
 spring:
+  main:
+    allow-bean-definition-overriding: true
```
bseeger, line 26: This allows me to inject the mocked address validation service. A good pattern for things like that, apparently.

bencalegari (approved):

PR #307, opened 2023-07-06
Fixes error in subflow where form submission data was not merged correctly 

bseeger:
addresses:  https://www.pivotaltracker.com/story/show/185546386

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -3,7 +3,12 @@ package formflow.library;
 import com.smartystreets.api.exceptions.SmartyException;
 import formflow.library.address_validation.AddressValidationService;
 import formflow.library.address_validation.ValidatedAddress;
-import formflow.library.config.*;
+import formflow.library.config.ActionManager;
+import formflow.library.config.ConditionManager;
+import formflow.library.config.FlowConfiguration;
+import formflow.library.config.NextScreen;
+import formflow.library.config.ScreenNavigationConfiguration;
+import formflow.library.config.SubflowConfiguration;
 import formflow.library.data.FormSubmission;
 import formflow.library.data.Submission;
 import formflow.library.data.SubmissionRepositoryService;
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -562,14 +561,21 @@ private String createFormActionString(String flow, String screen) {
   }
 
   private Map<String, Object> createModel(String flow, String screen, HttpSession httpSession, Submission submission) {
+    return createModel(flow, screen, httpSession, submission, null);
```
bencalegari, line 564: I'll make a change to just inline this reference rather than have the private method call an identically named version of itself.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -3,29 +3,13 @@
 import com.smartystreets.api.exceptions.SmartyException;
 import formflow.library.address_validation.AddressValidationService;
 import formflow.library.address_validation.ValidatedAddress;
-import formflow.library.config.ActionManager;
-import formflow.library.config.ConditionManager;
-import formflow.library.config.FlowConfiguration;
-import formflow.library.config.NextScreen;
-import formflow.library.config.ScreenNavigationConfiguration;
-import formflow.library.config.SubflowConfiguration;
+import formflow.library.config.*;
```
bseeger, line 6: I really don't like to do this.  It makes it hard to tell which items a file is using.

bseeger: A bit old, but... https://www.javadude.com/posts/20040522-import-on-demand-is-evil/

bseeger: If we are using google style guide (our syntax is from that), then they are not use in it. https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/main/java/formflow/library/ScreenController.java
bseeger: Ah, I hadn't check to see that there was only one other call to `createModel`... makes sense to do what you did then.

src/main/java/formflow/library/ScreenController.java
bencalegari: I have no strong opinions on it but IntelliJ does this automatically for me for some reason even though I have it turned off. Lemme try turning it off and putting it back.

bseeger: Ah, okay. You have to set the import limit to some high number - I don't think you can shut it off.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -39,634 +54,634 @@
...
-  ) throws NoHandlerFoundException {
-    log.info("getScreen: flow: " + flow + ", screen: " + screen);
-
-    var currentScreen = getScreenConfig(flow, screen);
-    Submission submission = submissionRepositoryService.findOrCreate(httpSession);
-    if ((submission.getUrlParams() != null) && (!submission.getUrlParams().isEmpty())) {
-      submission.mergeUrlParamsWithData(query_params);
-    } else {
-      submission.setUrlParams(query_params);
+    private final List<FlowConfiguration> flowConfigurations;
```
bseeger, line 57: Weird... Looks like it went to 4 spaces for tabs now.  Can you revert that?

bencalegari: Shouldn't the `PlatformFlavoredGoogleStyle` we all use apply this identically?
Screenshot 2023-07-06 at 2 19 09 PM

bseeger: weird, I thought so? 
Screenshot 2023-07-06 at 5 43 38 PM

bseeger: Not sure why yours is using 4 - the xml in our repo uses 2: https://github.com/codeforamerica/form-flow/blob/main/intellij-settings/PlatformFlavoredGoogleStyle.xml

bencalegari: Huh my bad I reimported it and it set properly, no idea how. Submitting a change to set that back.

PR #313, opened 2023-07-12
Adds a unit test for testing that subflow validation errors on input fields

bseeger:
Addresses testing of fix for: https://www.pivotaltracker.com/story/show/185546386

This adds a test of error handling on:
 * the first page of a subflow: the url `/flow/someflow/someSubflowPage/new` 
 * for an inner page of a subflow (one with established UUID): `/flow/someflow/someInnerSubflowPage/12301-eb2...`

I was noticing that subflow data was getting merged into the `inputData` at the top level in the `inputData` passed back to the page, via the Model. This separates that a bit more. It's now in the right spot in the `inputData` as well as in the `fieldData` passed as part of the Model to a page.

### DIFF
--- a/src/test/java/formflow/library/utilities/FormScreen.java
+++ b/src/test/java/formflow/library/utilities/FormScreen.java
@@ -29,8 +29,7 @@ public class FormScreen {
   }
 
   public boolean hasInputError(String inputName) {
-    Element element = html.select("input[name='%s'] ~ p.text--error".formatted(inputName))
-        .first();
+    Element element = html.select("[id^=\"%s-error-message\"]".formatted(inputName)).first();
     return element != null;
   }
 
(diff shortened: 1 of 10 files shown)


src/test/java/formflow/library/utilities/FormScreen.java
```diff
@@ -29,8 +29,7 @@ public boolean hasInputError() {
   }
 
   public boolean hasInputError(String inputName) {
-    Element element = html.select("input[name='%s'] ~ p.text--error".formatted(inputName))
-        .first();
+    Element element = html.select("[id^=\"%s-error-message\"]".formatted(inputName)).first();
```
bseeger, line 32: Not really sure why the `hasInputError` was checking something differently than the `getInputError`, but changed this to match that. I had a case where there was an error, and `getInputError` would fetch it correctly, but `hasInputError` said there wasn't an error.

bencalegari (approved):

PR #322, opened 2023-07-19
Learn how to localize 185485956

coltborg:
[#185485956](https://www.pivotaltracker.com/story/show/185485956)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Table of Contents
     * [PDF Generation](#pdf-generation)
     * [Sending Email](#sending-email)
         * [Mailgun](#mailgun)
+    * [Localization](#localization)
 * [How to use](#how-to-use)
     * [Configuration Details](#configuration-details)
         * [Environment Variables](#environment-variables)
@@ -2144,6 +2145,31 @@ error code and a corresponding error message as seen below.
 The error message thrown by mailgun is passed to the response object when an email fails. Common
 Mailgun error codes can be found
 [here](https://documentation.mailgun.com/en/latest/api-sending.html#error-codes).
+
+## Localization
+
+### Setting application properties
+
+Form flow library will read the languages set in an apps `application.properties`.
+
+```yaml
+# application.yaml
+form-flow:
+  languages: en, es
+```
+
+#### Language support
+
+TODO: talk about the language standard that's expected by Spring Boot
+
+### Supplying translation files
+
+Translation need to be added to `messages_xx.properties` files, where `xx` is the language the
+translations are for.
+
+### How Spring Boot handles localization
+
+### Mixpanel event from language selector
 
 # How to use
 
--- a/README.md
+++ b/README.md
@@ -2183,16 +2183,41 @@ the application's messages and can get overridden by message keys that exist in 
 To override a message from the library, use the same key in each of your messages files in your
 application.
 
-#### Using Tranifex
+#### Using [Transifex](https://www.transifex.com/)
 
 ##### File encoding in your application
 
-TODO: Mentioning that the message files are encoded for ISO-8859 (rather than the Spring Boot
-default UTF-8), in order to support Transifex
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2102,6 +2103,53 @@ The error message thrown by mailgun is passed to the response object when an ema
...
+
+Spring boot uses
+a [locale change interceptor](https://www.baeldung.com/spring-boot-internationalization#localechangeinterceptor)
+to intercept a `lang` query parameter (by default) to then change the locale
+and [sets a cookie](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/i18n/CookieLocaleResolver.html)
+to remember the locale.
+
+### Mixpanel event from language selector
+
+If your application is using mixpanel, and you want to keep track clicks into the language selector
```
bseeger, line 2149: I think it might be worth capitalizing Mixpanel (I think they capitalize it, but unsure). Also, a link to mixpanel.com, might be nice.

README.md
```diff
@@ -2102,6 +2103,53 @@ The error message thrown by mailgun is passed to the response object when an ema
 Mailgun error codes can be found
 [here](https://documentation.mailgun.com/en/latest/api-sending.html#error-codes).
 
+## Localization
+
+### Setting application properties
+
+Form flow library will read the languages set in an apps `application.properties`.
```
bseeger, line 2110: A little nit pick:  There's a few spots where you say "app" or "apps", I think it might read easier if the word "application" was used.

README.md
```diff
@@ -2102,6 +2103,53 @@ The error message thrown by mailgun is passed to the response object when an ema
...
+
+#### Language tag syntax
+
+Refer to
+the [language tag syntax](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#language_tag_syntax)
+to see what language you want to support.
+
+### Supplying translation files
+
+Translations need to be added to `messages_xx.properties` files in the `src/main/resources`
```
bseeger, line 2129: It might be good to specify/add `in the applications 'src/main/resources directory'`, so any reader realizes the configuration of this is all in their app and not the library.

README.md
```diff
@@ -2102,6 +2103,53 @@ The error message thrown by mailgun is passed to the response object when an ema
...
+directory, where `xx` is the language the translations are for.
+
+#### Messages in the library
+
+The library comes with [English](src/main/resources/messages-form-flow.properties)
+and [Spanish](src/main/resources/messages-form-flow_es.properties) messages. These load into the
+app messages and get override by message keys that exist in
```
bseeger, line 2136: here's a spot where "app" being "application's" would parse more easily (at least for my brain).

bseeger: Actually, in reading this more, it talks about the starter app and how it's configured.  It would be good to use that as an example, but maybe more this way: 

`These are loaded into the application's messages and can get overridden by message keys that exist in the application's message files.  Here's an example of how our Starter Application overrides some message strings...`

coltborg: I can't point to can example in the starter app, because we purposefully don't override library messages right now. But I added a sentence regarding this.

README.md
```diff
@@ -2145,6 +2146,71 @@ The error message thrown by mailgun is passed to the response object when an ema
...
+
+Spring boot uses
+a [locale change interceptor](https://www.baeldung.com/spring-boot-internationalization#localechangeinterceptor)
+to intercept a `lang` query parameter (by default) to then change the locale
+and [sets a cookie](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/i18n/CookieLocaleResolver.html)
+to remember the locale.
+
+### Mixpanel event from language selector
+
+If your application is using [Mixpanel](https://mixpanel.com/), and you want to keep track clicks
```
bseeger, line 2207: Is "track clicks" a Mixpanel thing, or should this be "track of clicks"?  If this is proper front end verbiage, then just ignore my comment.

PR #328, opened 2023-07-21
Upgrades 'org.springframework.boot' to version '3.1.2' and flyway to 9.21.0

bseeger:
We had previously held off on upgrading FlyWay, due to some changes that were incompatible with Spring Boot. Spring Boot made fixes in version 3.1.2, so now we can upgrade these libraries. 

https://github.com/spring-projects/spring-boot/issues/36029
https://github.com/flyway/flyway/issues/3694#issuecomment-1645499387

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     // Apply the java-library plugin for API and implementation separation.
-    id 'org.springframework.boot' version '3.1.1'
+    id 'org.springframework.boot' version '3.1.2'
     id 'io.spring.dependency-management' version '1.1.2'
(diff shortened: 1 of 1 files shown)


bseeger:
I do wonder if we should cut a release before merging this one...?

coltborg:
> I do wonder if we should cut a release before merging this one...?

No app is running latest snapshot are they?

But then again that would just mean waiting to automate a release Monday and then merging this.

coltborg (approved):
I tried this locally with `main` in starter app. Things worked as expected.

bencalegari (approved):

PR #334, opened 2023-07-26
Check flow exists in controller endpoints

spokenbird:


src/main/java/formflow/library/FileController.java
```diff
@@ -153,6 +164,9 @@ public ResponseEntity<?> upload(
 
       return ResponseEntity.status(HttpStatus.OK).contentType(MediaType.TEXT_PLAIN).body(newFileId.toString());
     } catch (Exception e) {
+      if (e instanceof ResponseStatusException) {
```
spokenbird, line 167: @bseeger I don't like this, but it was the only way I could think of to ensure a 404 gets thrown since this gobbles up exceptions of all types 🤔

src/test/java/formflow/library/controllers/ShowErrorStackTraceTest.java
```diff
@@ -10,18 +10,18 @@
...
+//    navigateTo("flow/inputs/asdf");
+//    String errorMessages = testPage.findElementsByClass("data-table").get(0).getText();
+//    String stackTrace = testPage.findElementsByClass("stack-trace").get(0).getText();
+//    assertThat(errorMessages).containsSequence("Timestamp:");
+//    assertThat(errorMessages).containsSequence("Status: 500");
+//    assertThat(errorMessages).containsSequence("Message: Could not find flow=inputs in templates");
+//    assertThat(errorMessages).containsSequence("Path: /flow/inputs/asdf");
+//    assertThat(errorMessages).containsSequence("Exception: java.util.NoSuchElementException");
+//    assertThat(stackTrace).containsSequence("java.util.NoSuchElementException: Could not find flow=inputs in templates");
+//  }
```
bencalegari, line 24: I'm assuming you meant to bring this back in?

src/test/java/formflow/library/controllers/ScreenControllerTest.java
```diff
@@ -47,6 +49,90 @@ public void setUp() throws Exception {
...
+    mockMvc.perform(MockMvcRequestBuilders.get("/flow/{flow}/{screen}/navigation", "flowThatDoesNotExist", "screen"))
+        .andExpect(status().isNotFound());
+  }
+
+  @Test
+  void getToScreenNavigationEndpointShouldReturn404WhenScreenDoesNotExist () throws Exception {
+    mockMvc.perform(MockMvcRequestBuilders.get("/flow/{flow}/{screen}/navigation", "testFlow", "screenThatDoesNotExist"))
+        .andExpect(status().isNotFound());
+  }
+
```
bencalegari, line 135: It'd be great for this to be two parameterized tests (one for posts and another for gets) with `@CsvSource` for the arguments since we're using the same methods for all of them, but it's not strictly necessary.

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -21,4 +29,27 @@ protected void saveToRepository(Submission submission, String subflowName) {
...
+    return flowConfigurations.stream().anyMatch(
+        flowConfiguration -> flowConfiguration.getName().equals(flow)
+    );
+  }
+
+  protected static void throwNotFoundError(String flow,String screen, String message) {
+    throw new ResponseStatusException(HttpStatus.NOT_FOUND, String.format("There was a problem with the request (flow: %s, screen: %s): %s",
+        flow, screen, message));
+
+  }
```
bencalegari, line 54: Glad to see these moved out of `ScreenController`! The big refactor I started did something similar for the first two methods so I'll have to reconcile that whenever I can get around to finishing it 😅

src/main/java/formflow/library/FileController.java
bencalegari: Does it not work to change 
```java
} catch (Exception e }) {
```
to
```java
} catch ResponseStatusException e }) {
```

spokenbird: I think the Exception e was meant as a sort of catch all for potential errors / issues with uploading to S3, although I'm not sure what the exact errors that would be thrown in that case are. I assume those would just get thrown as a 500 either way though?

src/main/java/formflow/library/FormFlowController.java
spokenbird: Would be interested to potentially pair on that when we can prioritize it.

src/test/java/formflow/library/controllers/ShowErrorStackTraceTest.java
spokenbird: Good catch, thanks!

PR #335, opened 2023-07-28
Validate max file size and file count server-side

bencalegari:
- FileTypeService -> FileValidationService. Seems like thats what this has ended up being now. 
- Switched /file-upload to MULTIPART_FORM_DATA_VALUE to clarify acceptable requests
- Added countBySubmission repository method

[#185691743]
[#185599082]

Co-authored-by: Chibuisi Enyia <cenyia@codeforamerica.org>


src/main/java/formflow/library/FileController.java
```diff
@@ -111,6 +120,10 @@ public ResponseEntity<?> upload(
         }
       }
 
+      if (userFileRepositoryService.countBySubmission(submission) >= maxFiles) {
```
bseeger, line 123: If we had 20 files all uploaded at the same time (through distinct but parallel requests), this might not be enough logic to prevent more than the max from being uploaded, given the test and set conundrum (first you test then you set, but there's a gap of time between testing and setting). But good enough for now.

src/main/java/formflow/library/file/FileValidationService.java
```diff
@@ -30,8 +30,9 @@
  */
 @Slf4j
 @Service
-public class FileTypeService {
+public class FileValidationService {
 
+  public static final int MB_IN_BYTES = 1024 * 1024;
```
bseeger, line 35: I was wondering if you thought about making this a long, since the filesizes will be long's (mostly). Don't feel strongly about it, but do worry about mixing data types in comparisons.  Java seems to do the right thing, but it's nice to keep datatypes aligned.

PR #336, opened 2023-07-28
Delete other jars when you jar

bencalegari:
- Once snapshots switch over, multiple jars sit in the build directory. This confuses starter-app locally since it can pick the wrong jar

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,13 @@ def sonatypeUsername = localEnv.getProperty("SONATYPE_USERNAME") ?: System.geten
 def sonatypePassword = localEnv.getProperty("SONATYPE_PASSWORD") ?: System.getenv("SONATYPE_PASSWORD")
 
 jar {
-    delete "build", "/libs/", "*.jar"
+    doFirst {
+        fileTree(dir: "$buildDir/libs", include: '*.jar').each { file ->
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -29,6 +29,7 @@ def sonatypeUsername = localEnv.getProperty("SONATYPE_USERNAME") ?: System.geten
 def sonatypePassword = localEnv.getProperty("SONATYPE_PASSWORD") ?: System.getenv("SONATYPE_PASSWORD")
 
 jar {
+    delete "build", "/libs/", "*.jar"
```
coltborg, line 32: I think the `clean` task deletes the whole `build` dir. Would that be a preferable way of doing it? Making `jar` call `clean` for us? Or is it better to be targeted a just `.jar`'s?

bencalegari: You're totally right, I didn't realize that. I also read that removing every `.jar` means you lose the benefits of [incremental builds](https://docs.gradle.org/current/userguide/incremental_build.html#sec:how_does_it_work) but there was a negligible difference when I ran `clean jar` compared to just `jar` after making a change. I've got a change that will just remove whichever jar doesn't match the current snapshot version which should have the same effect of removing anything you didn't intend to be there.

PR #341, opened 2023-08-01
Add logging to findOrCreate

tdooner:
We think this method is suspect for session splitting; let's add some
logging.

Co-Authored-By: Sree Prasad <sprasad@codeforamerica.org>
Co-Authored-By: Cypress Borg <cborg@codeforamerica.org>
Co-Authored-By: Ana Medrano Fernandez <amedrano@codeforamerica.org>

### DIFF
--- a/src/main/java/formflow/library/data/SubmissionRepositoryService.java
+++ b/src/main/java/formflow/library/data/SubmissionRepositoryService.java
@@ -34,7 +34,11 @@ public class SubmissionRepositoryService {
    * @return UUID of the saved submission
    */
   public UUID save(Submission submission) {
+    var newRecord = submission.getId() == null;
     UUID id = repository.save(encryptionService.encrypt(submission)).getId();
+    if (newRecord) {
+      log.info("created submission id: " + id);
+    }
     submission.setId(id);
     return id;
   }
--- a/src/main/java/formflow/library/data/SubmissionRepositoryService.java
+++ b/src/main/java/formflow/library/data/SubmissionRepositoryService.java
@@ -5,6 +5,8 @@ import java.util.ArrayList;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -85,9 +88,18 @@ public Submission findOrCreate(HttpSession httpSession) {
     var id = (UUID) httpSession.getAttribute("id");
     if (id != null) {
       Optional<Submission> submissionOptional = findById(id);
-      return submissionOptional.orElseGet(Submission::new);
+      if (submissionOptional.isEmpty()) {
```
bseeger, line 91: If we are going to add log messages, are there other ones that would add value as well?  

Maybe here: https://github.com/codeforamerica/form-flow/blob/main/src/main/java/formflow/library/FileController.java#L90. (this is the latest code, we'd have to find the spot in the 0.0.7 code).

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -85,9 +88,18 @@ public Submission findOrCreate(HttpSession httpSession) {
...
+        Submission newSubmission = new Submission();
+        log.info("findOrCreate created new submission: " + newSubmission.getId());
+        return newSubmission;
+      } else {
+        return submissionOptional.get();
+      }
     } else {
-      return new Submission();
+      Submission newSubmission = new Submission();
+      log.info("findOrCreate got no submission id from session, so created new submission: " + newSubmission.getId());
```
bseeger, line 101: note that the new submission id should be "null" since it hasn't been saved yet and won't get an id until it's in the DB (the DB assigns the id).

src/main/java/formflow/library/data/SubmissionRepositoryService.java
coltborg: I think our log from `findOrCreate` would also cover the `FilController`.

src/main/java/formflow/library/data/SubmissionRepositoryService.java
coltborg: We added another log in the `save` method to get that newly created id that show up `null` here.

PR #347, opened 2023-08-11
Fix cancel clearing and start investigating session cookie issue

coltborg:
* Remove cancelled file from array after it's been deleted
* Added a check to filter files from the session cookie that aren't in the db

### DIFF
--- a/src/main/java/formflow/library/FileController.java
+++ b/src/main/java/formflow/library/FileController.java
@@ -241,7 +241,6 @@ public class FileController extends FormFlowController {
       }
 
       httpSession.setAttribute(SESSION_USERFILES_KEY, dzFilesMap);
-      log.info("setAttribute {}", dzFilesMap);
 
       return new RedirectView(returnPath);
     } catch (Exception e) {
--- a/src/main/java/formflow/library/FileController.java
+++ b/src/main/java/formflow/library/FileController.java
@@ -94,9 +94,10 @@ public class FileController extends FormFlowController {
     log.info("POST upload (url: {}): flow: {}, inputName: {}", request.getRequestURI().toLowerCase(), flow, inputName);
     try {
       if (!doesFlowExist(flow)) {
-        throwNotFoundError(flow, null, String.format("Could not find flow with name %s in your application's flow configuration.", flow));
+        throwNotFoundError(flow, null,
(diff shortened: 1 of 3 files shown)


src/main/java/formflow/library/FileController.java
```diff
@@ -94,9 +94,10 @@ public ResponseEntity<?> upload(
     log.info("POST upload (url: {}): flow: {}, inputName: {}", request.getRequestURI().toLowerCase(), flow, inputName);
     try {
       if (!doesFlowExist(flow)) {
-        throwNotFoundError(flow, null, String.format("Could not find flow with name %s in your application's flow configuration.", flow));
+        throwNotFoundError(flow, null,
```
spokenbird, line 97: Is your styles/linter adding a line break?

src/main/java/formflow/library/FileController.java
```diff
@@ -203,7 +205,8 @@ RedirectView delete(
       HttpServletRequest request
   ) {
     try {
-      log.info("POST delete (url: {}): fileId: {} inputName: {}", request.getRequestURI().toLowerCase(), fileId, dropZoneInstanceName);
+      log.info("POST delete (url: {}): fileId: {} inputName: {}", request.getRequestURI().toLowerCase(), fileId,
```
spokenbird, line 208: Similar question as above here?

src/main/resources/META-INF/resources/webjars/form-flow/0.0.1/js/qunit-upload-tests.js
```diff
@@ -45,4 +45,4 @@ QUnit.module('Dropzone', function () {
 });
 QUnit.testDone(function () {
   document.querySelector("#uploadTest").className += " display-none";
-});
\ No newline at end of file
+});
```
spokenbird, line 48: Not actually sure what the difference is here?

src/main/java/formflow/library/FileController.java
coltborg: Yep, this was the linter.

src/main/java/formflow/library/FileController.java
coltborg: Yep, same, this was from the linter.

PR #353, opened 2023-08-21
Adds a section about the actuator endpoints

bseeger:
Adds some quick notes about the actuator, so people can understand the actuator and the risks it can expose us to.


MThorpester:
@bseeger This looks good to me!

PR #355, opened 2023-08-22
Update PDFBox to version 3.0.0

coltborg:
[#185861618](https://www.pivotaltracker.com/story/show/185861618)

Followed the [sparse 3.0 migration guide](https://pdfbox.apache.org/3.0/migration.html).

### DIFF
--- a/src/main/java/formflow/library/FileController.java
+++ b/src/main/java/formflow/library/FileController.java
@@ -21,6 +21,8 @@ import java.util.UUID;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.pdfbox.Loader;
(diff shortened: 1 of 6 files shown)


src/main/java/formflow/library/FileController.java
```diff
@@ -21,6 +21,8 @@
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.pdfbox.Loader;
+import org.apache.pdfbox.io.RandomAccessReadBufferedFile;
```
bseeger, line 25: This import doesn't appear to be used.

bseeger (approved):
Aside from one import that should come out, this looks good to me.

src/main/java/formflow/library/FileController.java
coltborg: Oh yes, will remove!

PR #369, opened 2023-08-31
Update flows-config-schema.json

spokenbird:
[#185752154]

### DIFF
--- a/intellij-settings/flows-config-schema.json
+++ b/intellij-settings/flows-config-schema.json
@@ -2,10 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Flows Configuration",
   "type": "object",
-  "required": [
-    "name",
-    "flow"
-  ],
+  "required": ["name", "flow"],
   "properties": {
     "name": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


bseeger:
Hi @spokenbird - can you explain how to test out this change? I applied the file, as described in our readme, and am not seeing it correct me. Not sure I'm doing it correctly. 
Thanks!

bseeger:
It started working now.  I am seeing the auto complete.

intellij-settings/flows-config-schema.json
```diff
@@ -43,14 +33,30 @@
                 }
               }
             },
-            "subflow": {
+            "onPostAction": {
               "type": "string",
-              "description": "For screen within a subflow, they need the name of the subflow defined."
+              "description": "A method name to run on a POST."
```
bseeger, line 38: I think we should fill this description out a little more.   Something along the lines of "This is is the name of an Action method that will be called when a POST occurs, before any general data processing happens."

I think you can say Action here, so folks can link this back to what an Action is in the docs.

intellij-settings/flows-config-schema.json
```diff
@@ -43,14 +33,30 @@
...
+            },
+            "crossFieldValidationAction": {
+              "type": "string",
+              "description": "A method name to to validate one field based on another field."
```
bseeger, line 42: "An Action that will be called after general validation is done on the fields, but before the data is saved to the database.  It is a chance to do any cross field validation necessary."

bseeger: I'd also fill the rest out some more as well.  If this is the only documentation that a SNE sees, then it's good if we give a little more context.

intellij-settings/flows-config-schema.json
```diff
@@ -43,14 +33,30 @@
...
             },
-            "beforeSave": {
+            "beforeSaveAction": {
```
bseeger, line 44: - before saving, but after validation
 - going point for

intellij-settings/flows-config-schema.json
spokenbird: Done!

bseeger (approved):
Looks good! Thanks for the updates!

PR #377, opened 2023-09-11
Removes <head> and <body> tags from fragments that don't need them.

bseeger:
Works towards resolving: https://www.pivotaltracker.com/story/show/185682485

[Starter app PR](https://github.com/codeforamerica/form-flow-starter-app/pull/210)

### DIFF
--- a/src/main/resources/templates/fragments/cardHeader.html
+++ b/src/main/resources/templates/fragments/cardHeader.html
@@ -1,7 +1,4 @@
-<!DOCTYPE html>
-<html th:lang="${#locale.language}" xmlns:th="http://www.thymeleaf.org">
-<body>
-<header
+<div
     th:fragment="cardHeader"
     th:with="hasSubtext=${!#strings.isEmpty(subtext)}"
     th:assert="${!#strings.isEmpty(header)}"
@@ -10,6 +7,4 @@
   <p id="header-help-message"
      th:if="${hasSubtext}"
      th:utext="${subtext}"></p>
-</header>
-</body>
-</html>
+</div>
--- a/src/main/resources/templates/fragments/cardHeader.html
+++ b/src/main/resources/templates/fragments/cardHeader.html
@@ -1,4 +1,4 @@
-<div
+<header
     th:fragment="cardHeader"
     th:with="hasSubtext=${!#strings.isEmpty(subtext)}"
     th:assert="${!#strings.isEmpty(header)}"
@@ -7,4 +7,4 @@
   <p id="header-help-message"
      th:if="${hasSubtext}"
      th:utext="${subtext}"></p>
-</div>
+</header>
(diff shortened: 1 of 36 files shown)


src/main/resources/templates/fragments/cardHeader.html
```diff
@@ -1,7 +1,4 @@
-<!DOCTYPE html>
-<html th:lang="${#locale.language}" xmlns:th="http://www.thymeleaf.org">
-<body>
-<header
+<div
```
bseeger, line 1: Wasn't sure if this was better off as a `<th:block>` or `<div>`.  Thoughts @coltborg?

coltborg: Here's my thought process on if to use a `<th:block>`:
* Do I want to an element rendered to the page?
  * If yes:
    * What is the most semantic element I can use? I'll try to use that.
    * If it's for display purposes or something technical that doesn't have useful information for a client, I'll pick a `<div>`.
  * If no:
    * I'll use `<th:block>`.

bseeger:
@coltborg another thought to go with this--- for the live template, there is a <templateSet> tag that we could wrap the live templates with so that they stop showing an error in our IDE (when editing them (multiple root issue)). I was trying to see if there was a different method to import them.

coltborg:
> @coltborg another thought to go with this--- for the live template, there is a `<templateSet>` tag that we could wrap the live templates with so that they stop showing an error in our IDE (when editing them (multiple root issue)). I was trying to see if there was a different method to import them.

ohhh cool, can you find any docs for `templateSet`? I'm only seeing [mention here for exporting](https://plugins.jetbrains.com/docs/intellij/template-support.html#export-the-live-template).

src/main/resources/templates/fragments/cardHeader.html
coltborg: For this one, I update the element and used `<header>` instead of `<div>`.

coltborg:
@bseeger digging into `templateSet` is outside of scope of the html tag ticket. I created another for this investigation and put it in the icebox - [#186018116](https://www.pivotaltracker.com/story/show/186018116).

bseeger:
I was looking at the live templates to see if any had these unnecessarily as well.  Should we just get rid of this template?
```
<template name="cfa:fragment" value="&lt;!DOCTYPE html&gt;&#10;&lt;html th:lang=&quot;${#locale.language}&quot; xmlns:th=&quot;http://www.thymeleaf.org&quot;&gt;&#10;&lt;body&gt;&#10; $CONTENT$&#10;&lt;/body&gt;&#10;&lt;/html&gt;" description="Base Thymeleaf fragment" toReformat="false" toShortenFQNames="true">
<variable name="CONTENT" expression="" defaultValue="" alwaysStopAt="true" />
<context>
  <option name="HTML" value="true" />
</context>
</template>
```

coltborg:
@bseeger, I removed the `cfa:fragment` live template 👍🏻

bseeger (commented):
LGTM!  I can't approve this as I worked on it (github won't let me).  I think this is good to merge though.

PR #380, opened 2023-09-14
Add interceptor to library

enyia21:


src/main/java/formflow/library/interceptors/DataRequiredInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+import org.springframework.util.AntPathMatcher;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+/**
+ * This interceptor prevents users from jumping to random pages in a flow.
+ */
+@Component
+@Slf4j
+@ConditionalOnProperty(name = "form-flow.session-continuity-interceptor.enabled", havingValue = "true")
+public class DataRequiredInterceptor implements HandlerInterceptor, Ordered {
```
bseeger, line 24: For naming: I can see that because we require both session and submission id, that it's data that's required, but maybe this would be more aptly named `SessionContinuityInterceptor` to make it immediately recognizable (to programmers) that it's related to the configuration parameter?

README.md
```diff
@@ -48,6 +48,11 @@ Table of Contents
     * [Sending Email](#sending-email)
         * [Mailgun](#mailgun)
     * [Localization](#localization)
+    * [Interceptors](#interceptors)
+        * [DataRequiredInterceptor](#datarequiredinterceptor)
+            * [Configuration](#configuration)
```
bseeger, line 53: Please only go three levels deep here, to keep the TOC shorter.   (keep the sections, just don't list them here beyond three levels. Thanks!)

README.md
```diff
@@ -2182,6 +2187,85 @@ in [our JavaScript](https://github.com/codeforamerica/form-flow/blob/9a1c6f7a253
 
 We send the whole url to Mixpanel, including any query parameters, such as `lang`.
 
+## Interceptors
```
bseeger, line 2190: We need a separate section about landmark pages, independent of this section.  This would be where we explain what landmark pages are intended for and catalog the ones that exist (right now, only the firstScreen, but I anticipate more soon).

src/main/java/formflow/library/interceptors/DataRequiredInterceptor.java
enyia21: I like it!  I'll get that changed

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+      throws Exception {
+    try {
+      String pathFormat = request.getRequestURI().contains("navigation") ? NAVIGATION_FLOW_PATH_FORMAT : FLOW_PATH_FORMAT;
+      Map<String, String> parsedUrl = new AntPathMatcher().extractUriTemplateVariables(pathFormat, request.getRequestURI());
+      String redirect_url = "/";
+      String screen = parsedUrl.get("screen");
+      String flow = parsedUrl.get("flow");
+      HttpSession session = request.getSession(false);
+      FlowConfiguration flowConfiguration = flowConfigurations.stream().filter(fc -> fc.getName().equals(flow)).findFirst()
+          .orElse(null);
```
mpidcock, line 54: This block of variables is a bit dense. Some suggestions for what I am used to:
* `redirect_url` seems like a class level variable, or maybe something that should be configured as part of each landmark config
* `screen` and `session` could be moved further down in the code to right before they are used - clearer to read and ensures they are only declared if needed
   * I might also inline `screen` since it is only used once
* `flow` is also only used once, but the line is rather long already. Here is the way I normally make `stream()` lines more readable:
```
FlowConfiguration flowConfiguration = flowConfigurations.stream()
          .filter(fc -> fc.getName().equals(parsedUrl.get("flow")))
          .findFirst()
          .orElse(null);
``` 
basically every chained dot method on the stream gets its own line

mpidcock (approved):
My comments are all codestyle and might differ from team norms so I won't demand that you change anything, but look it over at least

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
bseeger: I like to write streams this way as well, as I find it more readable.   I'll bring this to computer gardening to see if we can set a code style for this for the team.

bseeger:
@mpidcock and @enyia21  - Thanks for reviewing, Martha.  I know Martha is around on Monday. Not sure of @enyia21 schedule. I think it's worth working through some of these comments and updating the code based on the feedback.   It is well worth cleaning up some of the things, like exception handling, to make it more clear what's going on here.

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Sets the value of the interceptor to the highest integer value making it the last interceptor executed.
+   *
+   * @return Max Integer value.
+   */
+  @Override
+  public int getOrder() {
+    // Max value ensures that this interceptor is executed last.
```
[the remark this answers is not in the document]
enyia21: ✔️ Removed the comment.   We should rely on the javadoc to explain this method

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+
+      if (session.getAttribute("id") == null) {
+        log.error("A submission ID was not found in the session for request to {}. Redirecting to landing page.",
+            request.getRequestURI());
+        response.sendRedirect(redirect_url);
+        return false;
+      }
+
+      return true;
+    } catch (IllegalStateException e) {
```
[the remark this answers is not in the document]
enyia21: Yeah that feels weird as well.  I'm taking it out to see what breaks.   The prehandler still works the same in test

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+      return true;
```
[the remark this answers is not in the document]
enyia21: I removed the IllegalStateException for now.  I'm not sure if its necessary

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -40,61 +41,63 @@ public SessionContinuityInterceptor(List<FlowConfiguration> flowConfigurations)
...
+        .orElse(null);
 
-      if (flowConfiguration == null) {
+    if (flowConfiguration == null) {
+      return true;
+    }
+
+    if (flowConfiguration.getLandmarks() == null) {
+      throw new LandmarkNotSetException(
+          "The SessionContinuityInterceptor is enabled, but no 'landmarks' section has been created in the application's form " +
```
bseeger, line 65: Updated message here.

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -40,61 +41,63 @@ public SessionContinuityInterceptor(List<FlowConfiguration> flowConfigurations)
...
+              "flows configuration file.");
+    }
+
+    if (flowConfiguration.getLandmarks().getFirstScreen() == null) {
+      throw new LandmarkNotSetException(
+          "The SessionContinuityInterceptor is enabled, but a 'firstScreen' page has not " +
```
bseeger, line 71: Updated message here.

src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -40,61 +41,63 @@ public SessionContinuityInterceptor(List<FlowConfiguration> flowConfigurations)
...
+              "been identified in the 'landmarks' section in the application's form flows configuration file.");
+    }
+
+    String firstScreen = flowConfiguration.getLandmarks().getFirstScreen();
+
+    if (!flowConfiguration.getFlow().containsKey(firstScreen)) {
+      throw new LandmarkNotSetException(String.format(
+          "The form flows configuration file does not contain a screen with the name '%s'. " +
+              "Please make sure to correctly set the 'firstScreen' in the form flows configuration file 'landmarks' section.",
+          firstScreen));
```
bseeger, line 81: Updated message here.

src/test/java/formflow/library/interceptors/InterceptorMockMvcTest.java
```diff
@@ -78,7 +79,10 @@ void shouldErrorIfLandmarkIsNotSet() throws Exception {
         .andExpect(result -> {
           Exception resolvedException = result.getResolvedException();
           assertTrue(resolvedException instanceof LandmarkNotSetException, "Expected RuntimeException to be thrown");
-          assertEquals("You have enabled session continuity interception but have not created a landmark section in your applications flow configuration file.", resolvedException.getMessage());
+
+          assertEquals(
+              "The SessionContinuityInterceptor is enabled, but no 'landmarks' section has been created in the application's form flows configuration file.",
```
bseeger, line 84: In the tests I only updated error messages to match changes.

PR #388, opened 2023-09-26
Update date fieldset to use legend instead of label

spokenbird:
- Screen readers were not properly reading the label as it was not being associated with the fieldset. Legend is the more proper semantic HTML element for use with fieldsets.


bseeger (approved):
LGTM

coltborg (approved):
:shipit:

PR #390, opened 2023-09-27
Allow multiple flows w/in one session

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/186002002

Related changes to starter app in this PR: https://github.com/codeforamerica/form-flow-starter-app/pull/228

As part of this work, both the `UserFileRepositoryService` and `SubmissionRepositoryService` `save()` methods were changed to return an object versus the ID.   For the UserFile, it was because a test didn't have a way to set the id and was forcing us to do something in the main code we didn't like, just for the test. It seemed like a better move to rearchitect the code a little. 

Noting some useful resources:
 * https://thorben-janssen.com/spring-data-jpa-save-saveandflush-and-saveall/
 * https://www.codejava.net/frameworks/spring-boot/jpa-entitymanager-persist-vs-merge

Co-Authored-By: Chibuisi Enyia; cenyia@codeforamerica.org
Co-Authored-By: Alex Gonzalez; agonzalez@codeforamerica.org
Co-Authored-By: Martha Pidcock mpidcock@codeforamerica.org


src/main/java/formflow/library/ScreenController.java
```diff
@@ -474,10 +473,13 @@ ModelAndView navigation(
...
     log.info("GET navigation (url: {}): flow: {}, screen: {}", request.getRequestURI().toLowerCase(), flow, screen);
-    log.info("Current submission ID is :" + httpSession.getAttribute("id") + " and current Session ID is :" + httpSession.getId());
+    log.info(
+        "Current submission ID is :" + httpSession.getAttribute("id") + " and current Session ID is :" + httpSession.getId());
     // Checks if the screen and flow exist
     var currentScreen = getScreenConfig(flow, screen);
-    String nextScreen = getNextScreenName(submissionRepositoryService.findOrCreate(httpSession), currentScreen, null);
+    // TODO this change mirrors existing code, which doesn't account for putting a potentially new submission back into session... hmmm...
+    // this should probably fail in some way if Submission isn't already set, no? Can we get here with out an existing submission?
+    String nextScreen = getNextScreenName(findOrCreateSubmission(httpSession, flow), currentScreen, null);
```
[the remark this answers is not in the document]
bseeger: I changed it to this to try it out. Seems to work okay and gets rid of the TODO.

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -46,9 +61,110 @@ protected Boolean doesFlowExist(String flow) {
...
+  }
+
+  /**
+   * Returns the UUID of the Submission associated with the given flow.
+   *
+   * @param session The HttpSession the user is in
+   * @param flow    The flow to look up the submission ID for
+   * @return The submission id if it exists for the given flow, else null
+   */
+  public static UUID getSubmissionIdForFlow(HttpSession session, String flow) {
```
bseeger, line 100: How does this abstraction look? I liked putting this up here in case we change how we store the submissions again. If we do, then we will have less spots to change the logic of.

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -33,14 +32,14 @@ public SubmissionRepositoryService(SubmissionRepository repository, SubmissionEn
...
+    Submission savedSubmission = repository.save(encryptionService.encrypt(submission));
     if (newRecord) {
-      log.info("created submission id: " + id);
+      log.info("created submission id: " + savedSubmission.getId());
     }
-    submission.setId(id);
-    return id;
+    // straight from the db will be encrypted, so decrypt first.
+    return encryptionService.decrypt(savedSubmission);
   }
```
bseeger, line 43: Note that `save` now returns the Submission versus the id (as before).  This helped with some test code challenges and seems like a good thing to do overall.   Underneath, in the JPA, the object is either persisted or merged. If merged, the underlying save will return a different object. It seemed okay and useful to have this save return an object as well.

src/main/java/formflow/library/data/UserFileRepositoryService.java
```diff
@@ -27,8 +27,8 @@ public UserFileRepositoryService(UserFileRepository repository) {
    * @param userFile the uploadedFile to save, not null
    * @return UUID of the file
    */
-  public UUID save(UserFile userFile) {
-    return repository.save(userFile).getFileId();
+  public UserFile save(UserFile userFile) {
+    return repository.save(userFile);
   }
```
bseeger, line 32: Same thing as Submission - it seems better to return a new object, as it may have changed.

src/main/java/formflow/library/utils/UserFileMap.java
```diff
@@ -0,0 +1,64 @@
...
+ * then be shared with the client. Only include information that can be shared.
+ * </p>
+ */
+@Slf4j
+@Getter
+public class UserFileMap {
+
+  // flow -> inputName -> fileId -> file info
+  private Map<String, Map<String, Map<UUID, Map<String, String>>>> userFileMap;
+
```
bseeger, line 23: If you have any better ideas for this, I am all ears/eyes. But maybe we'd do that outside of this ticket.

src/test/java/formflow/library/utilities/AbstractBasePageTest.java
```diff
@@ -74,7 +74,7 @@ public void takeSnapShot(String fileWithPath) {
   }
 
   protected void uploadFile(String filepath, String dzName) {
-    testPage.clickElementById("drag-and-drop-box-" + dzName); // is this needed?
+    //testPage.clickElementById("drag-and-drop-box-" + dzName); // is this needed?
```
bseeger, line 77: I'll take this out as it doesn't seem to be needed (tests pass just fine w/o it).

src/main/java/formflow/library/FormFlowController.java
bseeger: There are three methods I moved to being here `getSubmissionIdForFlow`, `setSubmissionIdInSession`, and `findOrCreateSubmission`.

src/main/java/formflow/library/FileController.java
```diff
@@ -218,55 +205,59 @@ public ResponseEntity<?> upload(
...
    * @param returnPath           The path to the page that they came from
    * @param dropZoneInstanceName The drop zone instance used to get the user file name
+   * @param flow                 The name of the current (active) flow
    * @param httpSession          The current HTTP session
    * @return ON SUCCESS: Returns a RedirectView to the returnPath
    * <p>ON FAILURE: Returns a RedirectView to the 'error' page</p>
    */
   @PostMapping("/file-delete")
-  RedirectView delete(
+  public RedirectView delete(
```
spokenbird, line 214: why the public?

src/main/java/formflow/library/FileController.java
```diff
@@ -218,55 +205,59 @@ public ResponseEntity<?> upload(
...
-      HashMap<UUID, HashMap<String, String>> userFileMap = dzFilesMap.get(dropZoneInstanceName);
 
-      userFileMap.remove(fileId);
-      if (userFileMap.isEmpty()) {
-        dzFilesMap.remove(dropZoneInstanceName);
+      UserFileMap userFileMap = objectMapper.readValue((String) httpSession.getAttribute(SESSION_USERFILES_KEY),
+          UserFileMap.class);
+      if (userFileMap == null) {
+        log.error("User file map not set in session. Unable to update file information");
+        throw new IndexOutOfBoundsException("Session does not container user file mapping.");
```
spokenbird, line 257: container should be contain I think?

src/main/java/formflow/library/FileController.java
```diff
@@ -276,34 +267,44 @@ RedirectView delete(
...
+      @PathVariable String flow,
+      HttpSession httpSession,
       HttpServletRequest request
   ) {
     log.info("GET downloadSingleFile (url: {}): submissionId: {} fileId {}", request.getRequestURI().toLowerCase(), submissionId,
         fileId);
-    if (!submissionId.equals(httpSession.getAttribute("id").toString())) {
+
+    if (!submissionId.equals(getSubmissionIdForFlow(httpSession, flow).toString())) {
+      log.error("Submission ID handed in doesn't match the one associated with the file.");
```
spokenbird, line 290: Maybe include the ID handed in with the log?

Maybe something like "There was an attempt to download a file with submission ID {ID} that does not match the submission of the file being downloaded."

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -3,41 +3,56 @@
...
+    return saveToRepository(submission, null);
   }
 
-  protected void saveToRepository(Submission submission, String subflowName) {
+  protected Submission saveToRepository(Submission submission, String subflowName) {
     submissionRepositoryService.removeFlowCSRF(submission);
-    submissionRepositoryService.removeSubflowCSRF(submission, subflowName);
-    submissionRepositoryService.save(submission);
+    if (subflowName != null && !subflowName.isBlank()) {
+      submissionRepositoryService.removeSubflowCSRF(submission, subflowName);
```
spokenbird, line 42: Thanks for cleaning this up!

src/main/java/formflow/library/FormFlowController.java
spokenbird: Feels good to me!

src/main/java/formflow/library/utils/UserFileMap.java
spokenbird: Do we want to abstract the Map<UUID, <Map<String, String>> part into just UserFile?

src/main/resources/templates/fragments/fileUploader.html
```diff
@@ -271,7 +271,8 @@
       init: function () {
         window[dropzonePrefix + [[${inputName}]]] = this;
         var dzInstance = [[${inputName}]]
-        var documents = userFiles !== null ? userFiles[dzInstance] : [];
+        var flow = [[${flow}]]
+        var documents = userFiles?.userFileMap?.[flow]?.[dzInstance] || [];
```
spokenbird, line 275: I love that this works 😅

src/test/java/formflow/library/controllers/ScreenControllerJourneyTest.java
spokenbird, line 1: Confused why this is new? Thought this test already existed.

src/test/java/formflow/library/utilities/AbstractMockMvcTest.java
spokenbird, line 1: Confused what we are doing with the SessionAttrs?

src/main/java/formflow/library/FileController.java
bseeger: It's a public method so just being explicit about it, thought it's clearly not necessary. Other methods in this file are public as well (line 279).

src/main/java/formflow/library/utils/UserFileMap.java
bseeger: I can try this out. I was hesitant to because this is what our file uploader uses to load files on the page and it's different from what's in the UserFile object. That can be adjusted though.  I'll try an experiment.

src/main/resources/templates/fragments/fileUploader.html
bseeger: Yeah, this syntax is really nice.

src/main/java/formflow/library/FileController.java
bseeger: I wasn't sure how much information we wanted to expose here.  What was your philosophy on this when you created it?  I was thinking it might be handy to have the fileId in this msg as well.

src/main/java/formflow/library/FileController.java
bseeger: I've been wondering about the need to include method scope (public/protected/private).  Do you know and have documentation on the best practice for this using java 17? I couldn't find anything, so I just kept being explicit so I don't fall into a trap where the default scope is different then I think it is.

bseeger: Perhaps you already knew all of this...
In reading [this page](https://www.geeksforgeeks.org/public-vs-package-access-modifiers-in-java) that if you don't include the scope, the method is defaulted to public w/in its package.  It gives package-level access to this function. 

If we are explicit and put "public" there, then the method is accessible from anywhere, not just w/i the package. 

Given this, it doesn't make much of a difference if public is here or not - I think all things in our codebase should be able to access this method (not that they really need to, though).  

I was adding them to make the file more consistent, because some methods had it and some didn't.

src/test/java/formflow/library/utilities/AbstractMockMvcTest.java
bseeger: I removed the `sessionAttributes` in what turns out to be an easier flow for the code.  The tests add the flow info to the session and the session is sent along with any calls to our library interface.  It's a bit cleaner then passing around sessionAttributes, which did the same thing.

src/main/java/formflow/library/utils/UserFileMap.java
bseeger: I tried this out and am stuck on the thumbnail, which we take in, but don't store.  So if we pass back the UserFile, we won't have the thumbnail readily available with it.  
So then, do we start to store the thumbnail?  

I'd like to not do this right now but consider it, just outside of this ticket.

spokenbird: That seems totally reasonable. Part of the reason why we opted to use base64 encoded strings for the thumbs was so we didn't have to store the actual files. This change doesn't feel super necessary but could be worth considering down the road for quality of life when working with these more complex objects since it can be hard to maintain a mental model with all the nesting.

src/main/java/formflow/library/FileController.java
spokenbird: That's my understanding as well, that the absence of an access modifier indicates package-private, which as you stated gives access only to other classes within the same package.

I don't have much in the way of documentation but do see that Springs own example includes the public:
https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller.html

Maybe it makes sense to make this change across the board for consistency sake?

src/main/java/formflow/library/FileController.java
spokenbird: My thoughts were around providing the most information as to be helpful to SNEs if they are trying to debug a problem, or better understand what happened that lead to a file not being downloaded. In theory they should be the only ones seeing these logs.

In terms of exposing more or less, are you thinking about an outside actor gaining access to the logs?

bseeger: Yeah, an outside actor scenario.  I thought at one point folks were hesitant to include more file information in the logs, though I don't see it as a bad thing for us to do. 

I can include more information here, like the file id, so someone reading the log would have more to debug with.

spokenbird: There may be a question there for Dimitri. My thought was we didn't want to expose any PII in the logs (which things like original file names might include). Though I suppose if someone gained access to the logs they could see submission IDs and session IDs and potentially put them together to download information, though I think if someone gained log access we might have bigger problems?

src/main/java/formflow/library/FileController.java
bseeger: I agree - something we should do for consistencies sake.  I made a ticket to track this work: https://www.pivotaltracker.com/story/show/186243988

spokenbird (approved):

PR #391, opened 2023-09-27
Remove thumbnail dependency

coltborg:
We had removed much of the code. There was still a dependency that wasn't being used.


bseeger (approved):
I always love seeing libraries removed from dependencies. :)

PR #392, opened 2023-09-27
Change LandmarkNotSetException to extend exception

enyia21:
#186134647

### DIFF
--- a/src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
+++ b/src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
@@ -45,7 +45,7 @@ public class SessionContinuityInterceptor implements HandlerInterceptor, Ordered
    */
   @Override
   public boolean preHandle(HttpServletRequest request, @NotNull HttpServletResponse response, @NotNull Object handler)
-      throws Exception {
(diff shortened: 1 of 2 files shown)


src/main/java/formflow/library/interceptors/SessionContinuityInterceptor.java
```diff
@@ -45,7 +45,7 @@ public SessionContinuityInterceptor(List<FlowConfiguration> flowConfigurations)
    */
   @Override
   public boolean preHandle(HttpServletRequest request, @NotNull HttpServletResponse response, @NotNull Object handler)
-      throws IOException {
+      throws Exception {
```
bseeger, line 48: I think this might be clearer if it's `IOException, LandmarkNotSetException` to make it clear to folks exactly what they have to handle. 

Also the javadocs above need to include the two exceptions.

bseeger: To share more of my thinking. It's good practice to be specific when you can, hence using the specific exceptions rather than just saying it `throws Exception`. Technically that's true, but not very helpful to developers if they are quickly reading through the code.

bseeger (approved):
Looks good to me!

PR #393, opened 2023-09-28
Use flyway plugin

vrajmohan:


bseeger:
Hi @vrajmohan can you talk a little bit about why you think we should add this?  Is it to support migrations in the future, should we need them? Right now we do migrations via the java flyway library, so I'm not sure what we get with adding this.

bseeger:
So this would allow us to do database manipulations and setup migrations via gradle commands?   Sort of how ruby helps setup and run migrations?

Looks like we need to also add a section for the db credentials, or can you supply them when running the gradle command on the CLI?

bseeger:
Nice, I just plugged this data into my `gradle.properties` file and was able to run `flywayInfo` in Intellij. Pretty sweet!

```
flyway.user=starter-app
flyway.url=jdbc:postgresql://localhost:5432/starter-app
```

bseeger (approved):
Looks good!  Want to add this to the starter app as well? :)

PR #394, opened 2023-09-29
Moves setup.sh script into script directory and updates readme.

bseeger:


bseeger:
Mirroring the addition of the scripts directory in the starter app.  Not sure if we should do this though. Thoughts?

vrajmohan:
Do you expect many more scripts to be added? If so, IMO, it's worth creating a `scripts` directory. It may not be worth it for a couple of scripts.

bseeger:
I don't anticipate more scripts right now, so leaving it in the main repo is fine with me.  Wasn't sure if it was worth having the directory or not (sort of aligning the repos).

bseeger:
Closing as not really necessary.

PR #397, opened 2023-10-02
Improve check for database already existing

vrajmohan:
Lesson learned from form-flow-starter-app where a check for starter-app would succeed if start-app-test existed.


coltborg (approved):
Approved, same `grep` [as in the starter app](https://github.com/codeforamerica/form-flow-starter-app/commit/f72a4fd204e1026a290e9ed9ceb2be578e7afe13)

PR #398, opened 2023-10-03
Store useful attributes in the Mapped Diagnostic Context

vrajmohan:


README.md
```diff
@@ -2580,6 +2580,20 @@ There are spots in the templates where the `T` operator is used.
 
 #### Coming Soon!
 
+### Logging
+
+Form Flow adds the following attributes to the [Mapped Diagnostic Context](https://logback.qos.ch/manual/mdc.html):
+
+1. xForwardedFor
```
spokenbird, line 2587: Thinking about it further, we should probably add a brief description for each of these as others may not know what xForwardedFor is (like I didn't), etc.

spokenbird (approved):

PR #400, opened 2023-10-04
Add template resolver for a new location for uswds templates

coltborg:
* [Ticket #185770089](https://www.pivotaltracker.com/story/show/185770089)
* [Starter app PR](https://github.com/codeforamerica/form-flow-starter-app/pull/226)

**Question for reviewer:**
I had to create both `ThymeleafConfigurationUnloadedTest` and `ThymeleafConfigurationLoadedTest` in separate files so that I could load different application properties. Ideally I would like it to be in one file, maybe with nested classes, but I wasn't able to figure it out. If you have suggestions that would improve this, I'm very open to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1182,9 +1182,11 @@ The address fragment has two required parameters, `validate` and `inputName`.
 
 - `validate` is a boolean value that determines whether the address should be validated by Smarty
 - `inputName` is the name that will be associated with all of the above inputs by being used as a
-prefix in their input's name. For example, if the `inputName` is `homeAddress` then the corresponding
-inputs will be `homeAddressStreetAddress1`, `homeAddressStreetAddress2`, `homeAddressCity`, `homeAddressState`,
-and `homeAddressZipCode`.
+  prefix in their input's name. For example, if the `inputName` is `homeAddress` then the
+  corresponding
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -2347,6 +2350,17 @@ form-flow:
   path: 'name-of-file.yaml'
 ```
 
+#### Design System
+
+We are moving towards using a [custom theme](https://codeforamerica.github.io/uswds/dist/) of
```
spokenbird, line 2355: I might reword this. If I'm correct, it's actually setting the UWDS resolver to be used over the default, meaning that enabling it would turn on the USWDS resolver, but those templates will now replace the same ones from the Honeycrisp resolver? Meaning that turning this on will switch your application to using USWDS styles and components instead of Honeycrisp styles and components.

src/test/java/formflow/library/config/ThymeleafConfigurationUnloadedTest.java
spokenbird, line 1: I hate that it has to be done this way as two seperate tests, and I hate even more that I have to ask:

Do we want a third test that tests it defaults to Honeycrisp when no application property is set?

PR #410, opened 2023-10-12
Updates `iterationIsComplete` status

bseeger:
Addresses:  https://www.pivotaltracker.com/story/show/185779545


src/main/java/formflow/library/data/Submission.java
```diff
@@ -99,11 +105,20 @@ public Submission() {
...
+
+    List<Map<String, Object>> subflow = (List<Map<String, Object>>) inputData.get(subflowName);
+    List<Map<String, Object>> iteration = subflow.stream()
+        .filter(entry -> entry.get("uuid").equals(uuid))
+        .toList();
+
+    if (!iteration.isEmpty()) {
+      return iteration.get(0);
+    } else {
+      return null;
```
bseeger, line 120: Made this code less sensitive to failures, in the case the iteration was empty.

src/main/java/formflow/library/data/Submission.java
```diff
@@ -142,21 +157,36 @@ public void mergeFormDataWithSubflowIterationData(String subflowName, Map<String
...
+   * @return Boolean of true/false, maybe be null if the status is unset
+   */
+  public Boolean getIterationIsCompleteStatus(String subflow, String iterationUuid) {
+    Map<String, Object> iterationData = getSubflowEntryByUuid(subflow, iterationUuid);
+    if (iterationData == null) {
+      log.warn("Iteration completion status request, but iteration with id '{}' not found.", iterationUuid);
+      return null;
+    }
+
+    return (Boolean) iterationData.get(ITERATION_IS_COMPLETE_KEY);
```
bseeger, line 189: I added a setter and getter for this flag, so we could easily retrieve it from an iteration.

src/test/java/formflow/library/data/SubmissionTests.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Test
+  public void shouldRetrieveIterationIsCompleteStatusCorrectly() {
+    assertThat(submission.getIterationIsCompleteStatus("household", iterationUuid.toString())).isNull();
+    submission.setIterationIsCompleteStatus("household", iterationUuid.toString(), false);
+    assertThat(submission.getIterationIsCompleteStatus("household", iterationUuid.toString())).isFalse();
+    submission.setIterationIsCompleteStatus("household", iterationUuid.toString(), true);
+    assertThat(submission.getIterationIsCompleteStatus("household", iterationUuid.toString())).isTrue();
+  }
+}
```
bseeger, line 51: These are just simple unit tests to ensure that the setters and getters function as expected. We still need unit tests to ensure that calls via ScreenController function as expected where this flag is concerned.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -361,10 +363,16 @@ ModelAndView updateOrCreateIteration(
...
-    formSubmission.getFormData().put("iterationIsComplete", iterationIsComplete);
+
+    boolean isIterationComplete = !isNextScreenInSubflow(flow, submission, currentScreen, iterationUuid);
+
+    Boolean currentStatus = submission.getIterationIsCompleteStatus(currentScreen.getSubflow(), iterationUuid);
+    // if it is already true, leave it alone, otherwise set it
+    if (currentStatus == null || !currentStatus) {
+      submission.setIterationIsCompleteStatus(currentScreen.getSubflow(), iterationUuid, isIterationComplete);
+    }
   }
```
bseeger, line 376: I feel strongly that this logic of "only setting this flag if it's unset or false" stays in this class (versus being part of the setter/getting in Submission object). I don't think it should be in the Submission class. That class is concerned with a Submission, not with the business logic around a Submission (though we may have blurred the line elsewhere).

src/main/java/formflow/library/ScreenController.java
```diff
@@ -104,6 +104,10 @@ ModelAndView getScreen(
       submission.setUrlParams(query_params);
     }
 
+    if (uuid != null) {
+      updateIterationIsCompleteMarker(flow, uuid, submission, currentScreen);
+    }
+
```
bseeger, line 110: This is an attempt to update the marker when the SNE uses a GET to get to the end of a subflow. Quick testing is showing me it doesn't work, but I haven't debugged it yet. I half suspect that UUID is not included here, which maybe means that the check needs to be in a different spot??? Not sure.  

There is some code in the starter app that I was using to mimic the flow that this needs to work for:   https://github.com/codeforamerica/form-flow-starter-app/tree/iteration_is_complete_%23185779545

src/main/java/formflow/library/ScreenController.java
```diff
@@ -276,19 +281,8 @@ ModelAndView updateOrCreateIteration(
...
 
-    if (isNewIteration) {
-      // handle start iteration page, if new flow
-      HashMap<String, SubflowConfiguration> subflows = getFlowConfigurationByName(flow).getSubflows();
-      subflowName = subflows.entrySet().stream()
-          .filter(subflow -> subflow.getValue().getIterationStartScreen().equals(screen))
-          .map(Entry::getKey)
-          .findFirst()
-          .orElse(null);
-    }
```
mpidcock, line 290: While testing we unintentionally added two subflows with the start screens of the same name, which resulted in the `findFirst` grabbing the wrong one. After some investigation, we determined that since the subflowName is being set with `currentScreen.getSubflow()`, that value should always be present and this block is not needed.

README.md
```diff
@@ -246,6 +246,39 @@ asks the user to confirm their deletion before submitting the actual deletion re
...
+      "householdMemberFirstName": "Person",
+      "householdMemberRelationship": "Child",
+      "householdMemberRecentlyMovedToUS": "No"
+    }
+  ],
+```
+Note that all information submitted for a single loop (iteration) through your subflow will show up within a single array index.
+The index includes a `uuid` field, which is a unique identifier for that iteration within the subflow. 
+
+#### Completed iterations
```
enyia21, line 275: Are we capturing that when the last screen is a GET requests out of a subflow,  iterationIsComplete is set to true?

mpidcock: Made a small update to clarify

src/main/java/formflow/library/ScreenController.java
```diff
@@ -361,10 +354,13 @@ ModelAndView updateOrCreateIteration(
     return new ModelAndView(viewString);
   }
 
-  private void setIterationIsComplete(String flow, String iterationUuid, FormSubmission formSubmission, Submission submission,
+  private void updateIterationIsCompleteMarker(String flow, String iterationUuid, Submission submission,
       ScreenNavigationConfiguration currentScreen) {
-    Boolean iterationIsComplete = !isNextScreenInSubflow(flow, submission, currentScreen, iterationUuid);
-    formSubmission.getFormData().put("iterationIsComplete", iterationIsComplete);
+    // Set iterationIsComplete to true if the next screen is not in the subflow
+    boolean isIterationComplete = !isNextScreenInSubflow(flow, submission, currentScreen, iterationUuid);
```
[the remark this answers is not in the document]
mpidcock: Ended up refactoring this code overall to be less repetitive/simplified. It involved moving the logic for next screen in subflow & is iteration complete into one place within the navigation endpoint, which is now used by both the GET and POST within a subflow. Thanks for highlighting this section!

PR #411, opened 2023-10-12
Updated javadocs for MailgunEmailClient and EmailClient

enyia21:
#186242068

### DIFF
--- a/src/main/java/formflow/library/email/EmailClient.java
+++ b/src/main/java/formflow/library/email/EmailClient.java
@@ -4,7 +4,7 @@ import java.io.File;
 import java.util.List;
 
 /**
- * Requires sendEmail methods for any class implementing EmailClient.
+ * An interface that can be implemented by objects to send email.
  * @param <T> A Generic that can be set as the return type for the sendEmail method
  */
 public interface EmailClient<T> {
--- a/src/main/java/formflow/library/email/EmailClient.java
+++ b/src/main/java/formflow/library/email/EmailClient.java
@@ -3,8 +3,15 @@ package formflow.library.email;
 import java.io.File;
 import java.util.List;
 
+/**
+ * Requires sendEmail methods for any class implementing EmailClient.
+ * @param <T> A Generic that can be set as the return type for the sendEmail method
+ */
 public interface EmailClient<T> {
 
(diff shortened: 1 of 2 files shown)


src/main/java/formflow/library/email/MailgunEmailClient.java
```diff
@@ -19,9 +19,21 @@
...
+   * String storing the Mailgun domain
+   */
   private final String mailgunDomain;
+  /**
+   * Mailgun messages api object
+   */
   private MailgunMessagesApi mailgunMessagesApi;
+  /**
+   * Boolean variable sets requireTLS to true.
+   */
```
bseeger, line 36: Do we need to document private variables? I doubt these show up in the javadocs...?

bseeger: Not that it hurts to document the, but I don't think we _have_ to.

bseeger: Do you have any thoughts on this?  I'm fine keeping it in, just wondering if you knew of a best practice around this?

enyia21: Documenting private variables is not included in the javadocs warnings for missing comments

src/main/java/formflow/library/email/EmailClient.java
```diff
@@ -3,8 +3,15 @@
...
+/**
+ * Requires sendEmail methods for any class implementing EmailClient.
+ * @param <T> A Generic that can be set as the return type for the sendEmail method
+ */
 public interface EmailClient<T> {
 
+  /**
+   * A boolean variable required for any EmailClient
+   */
   Boolean requireTls = true;
```
bseeger, line 15: This inspired conversation about EmailClient being an interface and default scope for this variable.  When there is a variable in an Interface, it is by default "public Static final", which is obviously what we don't want here.  We are creating a ticket to rethink the logic here.

bseeger: https://www.pivotaltracker.com/story/show/186251813

bseeger (approved):
Looks good to me!

PR #416, opened 2023-10-17
Adds `postfix` capabilities to both Money and Number Inputs.

bseeger:
Co-Authored-By: Lauren Kemperman; lkemperman@codeforamerica.org

We needed this ability for a few screens in LA Assister and it seems handy to just have this in the library.


src/main/resources/templates/fragments/inputs/number.html
```diff
@@ -17,19 +17,22 @@
...
+             th:name="${inputName}"
+             th:placeholder="${placeholder}"
+             th:attr="
+              title=${hasTitle ? title : #messages.msg('general.inputs.number')},
+              pattern=${hasPattern ? pattern : '^\d*(\.\d{0,2})?$'},
+              aria-describedby=${hasHelpText ? inputName + '-help-text' : ''},
+              aria-labelledby=${hasAriaLabel ? ariaLabel : ''},
+              aria-invalid=${hasError}"
+             th:value="${fieldData.getOrDefault(inputName, '')}">
+      <div th:if="${postfix != null}" th:text="${postfix}" class="text-input-group__postfix"></div>
```
coltborg, line 33: When using voiceover, the `postfix` text is not being described. Idk if this is a blocker - @spokenbird?


https://github.com/codeforamerica/form-flow/assets/9101728/35d6ec12-4787-40a1-93d9-61b23766ef80

bseeger: Nice catch! Thanks for thinking of this test.

coltborg (approved):
We discussed this in standup, the prefix also does not get read by a screen reader, so we will be addressing those issues at the same time.

PR #421, opened 2023-10-23
Add method to user file map

bseeger:
This adds a convenience method to get the user files from the UserFileMap, based on `flow` and `inputName`. 
We have the ability to call `getUserFileMap()`, but that provides the entire map and then the user (in this case a utility class supporting a thymeleaf template) will have to do the digging through it.  Seems like a nice thing for use to encapsulate the complexity of that data structure. 

This also fills in Javadoc information for the UserFileMap class as well as adds a quick test for the new method.


coltborg (approved):
Looks good, thank you for also adding javadocs!

PR #424, opened 2023-10-25
Adds migration and column for `doc_type_label` and fixes virus scanning logic

lkemperman-cfa:
* Adds logic to create space to set the doc type for a file:
  * Adds `doc_type_label` to `user_files` table. 
  * Adds `form-flow.uploads.default-doc-type-label` so one can set the default document type if they want.
  * Updates the readme to include this new field and how to set the default.  
* Adds `@DynamicField` annotation and separates annotations from their validators (slight package reorg)
* Fixes an error where files were marked incorrectly if virus scanning was disable. File were getting marked as `true` when scanning was turned off.  The NoOpVirusScanner is used when the virus scanning is disabled, so it was marking the record inaccurately.  We revamped the logic around this and made the check a little clearer. 

You can test this ticket out by using this [branch in the LA Doc Uploader](https://github.com/codeforamerica/la-doc-uploader/tree/update-file-uploader-pages).


README.md
```diff
@@ -1327,24 +1340,55 @@ The resulting file information will be stored in the database in two places:
 1. the `submissions` table
 2. the `user_files` table
 
-The saved JSON is store in the `submissions` table with the field name being the key.
+The `submissions` table is where the file is associated with a submission as part of the
```
bseeger, line 1343: This where the real changes start. Above is just formatting tweaks.

README.md
```diff
@@ -1315,7 +1328,7 @@ once the upload is complete it will become a delete link.
 If the file has an extension of .jpg, .png, .bmp or .gif, DropZone will create a thumbnail for that
 image and display it on the screen.
 
-For files with extensions other than the ones listed (like tif files and various document formats),
+For files with extensions other than the ones listed (like .tif files and various document formats),
```
bseeger, line 1331: https://github.com/codeforamerica/form-flow/pull/424

src/main/java/formflow/library/ValidationService.java
```diff
@@ -102,6 +104,13 @@ private Map<String, List<String>> performFieldLevelValidation(String flowName, F
         key = key.replace("[]", "");
       }
 
+      String originalKey = key;
+
+      if (key.contains("_wildcard")) {
```
spokenbird, line 109: Wondering if _Nth would be more appropriate here? Where Nth represents some number in an unknown total?

Not a big deal, _wildcard works, but Nth came to mind.

src/main/resources/db/migration/V2023.10.25.08.53.44__create_doc_type_label_for_user_files.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_files
+    ADD COLUMN doc_type_label VARCHAR(255) DEFAULT '${user_file_doc_type_default_label}';
```
spokenbird, line 2: Should we add a test to ensure the default is set when used?

README.md
```diff
@@ -1327,24 +1340,55 @@ The resulting file information will be stored in the database in two places:
...
+
+The `user_files` table holds information about files a user uploaded as part of their submission.
+The table includes:
+
+* `file_id` - the ID of the file (UUID)
+* `filesize` - size of file, in bytes
+* `submission_id` - ID of the submission the file relates to
+* `created_at` - time file was uploaded
+* `original_name` - original name of the file at the time of upload
+* `doc_type_use` - the document type use label for the file
```
spokenbird, line 1370: Confused about the difference between doc_type_use and doc_type_label?

bseeger: Oh, that's a typo - it should be `doc_type_label` - I'll fix it.  We are hoping that the name `doc_type_label` is unique enough that folks don't think it's technical metadata like mimetype.

src/main/java/formflow/library/ValidationService.java
bseeger: We need something we could match on easily - the Nth idea would be hard to match short of regex and w/o a prefix to match one.  However, what you're suggesting sounds like a way to track the **number** of "dynamic" fields, if need be.  I could see how that could be handy in some situations.  

Like what about: "_multiField_NN_", where N is a number (0-9). 

Does make me think that the "_wildcard" should, at the very least, be a static string const.

src/main/resources/db/migration/V2023.10.25.08.53.44__create_doc_type_label_for_user_files.sql
bseeger: I didn't address this one, but can look at it.

bseeger: I tried to add a test for this but am seeing the test db act differently then the live db. Namely defaults and required fields are not getting set in test. This makes it so I cannot test the default value of the doc type field. I did create tests and check it in, but it's disabled because the db isn't correct.

src/main/java/formflow/library/inputs/FieldDesignation.java
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * A class to capture any string constants that might apply to fields in the inputs class.
+ */
+public class FieldDesignation {
+
+  /**
+   * A string used to indicate if an input field is a dynamic field. If this string is found in a field's name in a form
+   * submission, then it's likely the field is a dynamic field.
+   */
+  public static final String DYNAMIC_FIELD_MARKER = "_wildcard_";
```
coltborg, line 12: @bseeger what's the difference between this file and `UnvalidatedField.java`? Both are a class with defined strings in them that some fields might have. Though I think the dynamic fields still want to go through validation service.

Can the two files be combined?

coltborg:
One last comment and there's a conflict in `README.md`, then this PR is good to go.

src/main/java/formflow/library/inputs/FieldDesignation.java
bseeger: Yes, I think they totally could.  The naming of `UnvalidatedField` doesn't seem to encompass what we're trying to do here, since the dynamic fields are validated, just that the marker and trailing suffix are taken off first.   I'll tinker with naming and see what I can come up with for this.  Let me know if you have an ideas.

bseeger: I changed this around - let me know what you think @coltborg

coltborg (approved):
Comments addressed, approved.

PR #425, opened 2023-10-25
Add a test for going from a regular POST into a subflow

spokenbird:
- Adds a test to prevent regression with regards to https://github.com/codeforamerica/form-flow/commit/c05abf334c0e1bc654cd661735d9c9c39c1e9863


src/test/resources/templates/testSubflowLogic/subflowAddItem.html
```diff
@@ -9,8 +9,8 @@
       <main id="content" role="main" class="form-card spacing-above-35">
         <th:block
             th:replace="~{fragments/cardHeader :: cardHeader(header='Subflow Page', subtext='This is a page within a subflow')}"/>
-        <!-- Test that the currentSubflowItem is passed in by the controller: -->
-        <div th:text="${currentSubflowItem.firstNameSubflow}"></div>
```
spokenbird, line 13: I'm not sure what this was doing exactly but firstNameSubflow was null and I couldn't figure out what would have been setting it in the first place.

bseeger: I wonder if that was debug that did get taken out?

bseeger: Should that `div` even be there?  Not exactly sure what it's doing either.

spokenbird: I think you might be right, I'll take it out all together.

coltborg:
@spokenbird, when your back, what's the status of this PR? Is is ready to be reviewed?

src/test/resources/templates/testSubflowLogic/subflowAddItem.html
mpidcock: @spokenbird Did you decide to remove this?

PR #427, opened 2023-10-26
Adding icons from design library and fixing clip id conflict

sree-cfa:
See [FFB Design Library](https://www.figma.com/file/Di2IIwIta6Cyk7EfTcOvKp/Form-Flow-Builder-Design-Library?type=design&node-id=1-11504&mode=design&t=7GBY3qyFw6Xzd5zj-0)


src/main/resources/templates/fragments/icons.html
```diff
@@ -780,7 +945,7 @@ <h1>Icons</h1>
                   stroke-linecap="round" stroke-linejoin="round"/>
           </g>
           <defs>
-            <clipPath id="clip0">
```
coltborg, line 783: Why did `clip0` need to be updated? Is there a `clip0` in a different icon?

src/main/resources/templates/fragments/icons.html
```diff
@@ -8,6 +8,124 @@ <h1>Icons</h1>
...
+          <path d="M36.5 50.4053V57.479" stroke="black" stroke-width="2" stroke-linejoin="round"/>
+          <path d="M36.5 58.6575L36.5 61.0154" stroke="black" stroke-width="2" stroke-linejoin="round"/>
+        </svg>
+
+      </td>
+    </tr>
+    <tr>
+      <td>filledLock</td>
+      <td>
+        <svg th:fragment="filledLock" width="19" height="30" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
coltborg, line 101: Some of these icons look like they're from Google Material Icons, which comes in directly from [Honeycrisp](https://honeycrisp.herokuapp.com/cfa/styleguide#atoms-icons).

new-icons


Some don't have a class, like the `filledLock` or `lock`. But you could use character codepoint, for example:

```
<i class="icon-">&#xe897;</i>
```

Which would give you:
Screenshot 2023-10-26 at 4 14 22 PM


The full list of [codepoint reference are here](https://github.com/google/material-design-icons/blob/f5f56570741833bdd36463f1f1b6b7d4edd3f9c1/font/MaterialIconsOutlined-Regular.codepoints), linked from Honeycrisp. 

The format would be `<i class="icon-">&#INSERT_CODEPOINT_HERE;</i>`.

If there's an icon that can't be accessed via codepoint, then I'd be ok with having them here. But we're already loading the icon via a material icon font and this could be a redundant home for them.

src/main/resources/templates/fragments/icons.html
sree-cfa: Yeah, clip0 appears a few times in here. We ran into an issue trying to display it on the same page

src/main/resources/templates/fragments/icons.html
sree-cfa: 🤯

coltborg (approved):
Approved

PR #434, opened 2023-11-06
Updates fragments with tweaks from LA projects

bseeger:
We are trying to move small tweaks into the FFB to make them useful for other projects.


src/main/resources/templates/fragments/inputs/submitButtonSecondary.html
```diff
@@ -0,0 +1,9 @@
+<th:block
```
bseeger, line 1: This one I'm not sure if we should add or not. What do you think?  Is it worth having a separate one just for the secondary color or should we have a way to pass in style, which seems odd as well?

spokenbird: Doesn't hurt, makes it easier to use since the first hard codes the button--primary class.

spokenbird:
@bseeger these are great changes to pull in. For the template changes for the additional fields on the reveal and the address components I think we should add documentation to clue people into the fact that these optional fields are available to them?

spokenbird:
@bseeger just following up on this, do you have the time to document the changes?

bseeger:
I'll add some documentation to this later today.

bseeger:
@spokenbird I added some documentation - I didn't have a chance to double check it yet though.

bseeger:
@spokenbird I think this is all set now. What do you think?

spokenbird (approved):

PR #437, opened 2023-11-06
Create configuration for locking the Submission for a flow after it is submitted

spokenbird:


spokenbird:
This PR adds the ability to configure locking down a submission such that it can no longer be edited after the submission has been submitted (which adds a submittedAt time-stamp to the submission object). 

This is achieved by adding configuration to the `application.yaml` file like so:

```yaml
 lock-after-submitted:
   - flow: ubi
     redirect: success
```

and by adding an `afterSubmitPages` section to your `flows-config.yaml` file's `landmarks` section for the flow you are locking like so:

```yaml
landmarks:
  firstScreen: howThisWorks
  afterSubmitPages:
    - nextSteps
    - success
```

The changes in this PR will cause a user who was already submitted their application to be redirected to the `redirect` indicated in the `lock-after-submitted` section, **unless** the page they are accessing is one of those listed in the `afterSubmitPages`.

Note that this PR does not actually lock down the Submission object, or any rows in the Submission DB table, but rather just prevents the user from POSTing any data or GETing any screen not indicated in `afterSubmitPages`.

Additional changes in this PR include a refactor of the `FlowsConfigurationFactory` and the `SessionContinuityInterceptor`.

The `FlowsConfigurationFactory` will now validate flows before adding them to application context. It does this by checking if the flow is configured to be disabled, and if so, it will not add it to the application context. Furthermore, it will now validate that landmarks have been correctly set if both if a flow is configured to lock it's submissions after submit, or if a flow has enabled the session continuity interceptor. 

If submission locking is enabled for a flow, the `FlowsConfigurationFactory` will throw an error when validating the `FlowConfiguration` for that flow if it does not find an `afterSubmitPages` or `landmarks` section.

It will do the same when the `SessionContinuityInterceptor` is enabled but no `landmarks` `firstScreen` has been configured for the given flow. 

These errors occur during the applications build time and as such will fail when the SNE attempts to run an application that has been misconfigured, with error messages alerting them to the issue and steps needed to remediate the problem.

bseeger:
Hi - I can look at this early next week.

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -167,4 +177,24 @@ protected void setSubmissionInSession(HttpSession session, Submission submission
...
+
+  /**
+   * Determines if the user should be redirected due to a locked submission.
+   *
+   * @param flowName     The name of the flow.
+   * @param screen       The current screen name.
+   * @param submission   The submission object.
+   * @return true if the user should be redirected, false otherwise.
+   */
+  public boolean shouldRedirectDueToLockedSubmission(String flowName, String screen, Submission submission) {
```
spokenbird, line 189: Flow can just be grabbed from the flow object

src/test/java/formflow/library/config/FlowsConfigurationFactoryTest.java
```diff
@@ -0,0 +1,104 @@
...
+@TestPropertySource(properties = {"form-flow.session-continuity-interceptor.enabled=true"})
+public class FlowsConfigurationFactoryTest {
+
+  private FlowsConfigurationFactory flowsConfigurationFactory;
+  private FormFlowConfigurationProperties formFlowConfigurationProperties;
+  
+  private SessionContinuityInterceptorConfiguration sessionContinuityInterceptorConfiguration;
+  private ClassPathResource classPathResource;
+  
+  String AFTER_SUBIT_MISCONFIGURATION_ERROR = "You have enabled submission locking for the flow testFlow but the afterSubmitPages landmark is not set in your flow configuration yaml file.";
```
spokenbird, line 26: Fix typo `SUBIT`

src/test/java/formflow/library/config/FlowsConfigurationFactoryTest.java
```diff
@@ -0,0 +1,104 @@
...
+    FlowConfiguration flowConfig = new FlowConfiguration();
+    HashMap<String, ScreenNavigationConfiguration> screenNavigationConfigurations = new HashMap<>();
+    screenNavigationConfigurations.put("testFirstScreen", new ScreenNavigationConfiguration());
+    flowConfig.setFlow(screenNavigationConfigurations);
+    flowConfig.setName("testFlow");
+
+    // The factory should throw an exception when it tries to validate the flow since there is no landmarks section but session continuity interception is enabled
+    FlowConfigurationException errorThrownForNoLandmarkSectionAtAll = assertThrows(FlowConfigurationException.class,
+        () -> flowsConfigurationFactory.validateLandmarksFirstScreen(flowConfig));
+    assertThat(errorThrownForNoLandmarkSectionAtAll.getMessage()).isEqualTo(FIRST_SCREEN_MISCONFIGURATION_ERROR);
```
spokenbird, line 76: Update these to use assertEquals

PR #446, opened 2023-11-15
Pin flyway to 9.x versions

vrajmohan:
10.0 seems to have introduced breaking changes.
See https://github.com/flyway/flyway/issues/3780

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
     id 'jacoco'
     id 'signing'
     id 'io.github.gradle-nexus.publish-plugin' version '1.3.0'
-    id "org.flywaydb.flyway" version "10.0.1"
+    id "org.flywaydb.flyway" version "[9.+,10.0["
 }
 
 jacoco {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -75,7 +75,7 @@ dependencies {
     implementation 'org.jetbrains:annotations:24.0.1'
     implementation 'io.hypersistence:hypersistence-utils-hibernate-60:3.6.1'
     implementation 'javax.persistence:javax.persistence-api:2.2'
-    implementation 'org.flywaydb:flyway-core:9.22.3'
+    implementation 'org.flywaydb:flyway-core:[9.+,10.+['
```
coltborg, line 78: @vrajmohan - The goal is to allow any latest version of `9.x`, and disallow any version of `10.x` right? 

Is it meaning full that on line 12 it is `10.0[` version here on line 78 it is `10.+[`?

vrajmohan: Yes, the goal is to allow any latest version of 9.x, and disallow any version of 10.x.

It is based on my reading of https://docs.gradle.org/current/userguide/single_versions.html.
> The symbol ] can be used instead of ( for an exclusive lower bound, and [ instead of ) for exclusive upper bound. e.g ]1.0, 2.0[

coltborg (approved):
Approved!

PR #448, opened 2023-11-16
Remove JPA DDL schema creation 186402340

vrajmohan:


src/test/resources/application-test.yaml
```diff
@@ -45,9 +45,6 @@ spring:
   messages:
     encoding: ISO-8859-1
     basename: messages, messages-form-flow
-  jpa:
-    hibernate:
-      ddl-auto: create-drop
```
bseeger, line 50: Does this mean our DB data stays around for the next run?  I thought this aided in clearing out the db at the end of tests, but I could be wrong.

src/test/java/formflow/library/repository/UserFileRepositoryServiceTests.java
```diff
@@ -83,7 +73,12 @@ void shouldUserDocTypeSetInUserFileBuilderCall() {
         .docTypeLabel("BirthCertificate")
         .build();
 
-    UserFile savedFile = userFileRepositoryService.save(testFile);
+    UserFile savedFile = saveAndReload(testFile);
     assertThat(savedFile.getDocTypeLabel()).isEqualTo("BirthCertificate");
   }
+
+  private UserFile saveAndReload(UserFile testFile) {
+    UserFile savedFile = userFileRepositoryService.save(testFile);
```
bseeger, line 81: Save should have returned the updated object - either having updated the one handed in or a new object that's the result of an internal merge.  But you're not seeing the new object here?  But things function normally outside of tests?

src/test/resources/application-test.yaml
vrajmohan: The clearing out is handled by `org.springframework.boot.autoconfigure.flyway.FlywayMigrationStrategy`.

src/main/java/formflow/library/data/UserFile.java
```diff
@@ -30,6 +31,7 @@
 
 @Entity
 @Table(name = "user_files")
+@DynamicInsert
```
bseeger, line 34: This is interesting - do we need this on the submission table as well?

src/test/java/formflow/library/repository/UserFileRepositoryServiceTests.java
vrajmohan: Yes, that is what I am seeing, contrary to expectation. It _could_ be something else we are doing wrong.

src/main/java/formflow/library/data/UserFile.java
vrajmohan: We need it if we are relying on the database to supply defaults.

vrajmohan:
Converted to draft PR as there are 2 other tests failing because we are running against the _real_ schema.

bseeger:
Thanks for addressing this @vrajmohan. My only concern has nothing to do with your work and everything to do with the question of why test is behaving differently then the live code base.

src/test/resources/application-test.yaml
vrajmohan: It is actually https://github.com/codeforamerica/form-flow/blob/9c189f9ba6a22a66a03496eab42697c7063fc0c6/src/test/java/formflow/library/config/FlywayConfiguration.java#L8

src/test/java/formflow/library/repository/UserFileRepositoryServiceTests.java
vrajmohan: You can see a similar pattern already in use here - https://github.com/codeforamerica/form-flow/blob/main/src/test/java/formflow/library/repository/SubmissionRepositoryServiceTest.java#L60-L62

bseeger (approved):
What you have LGTM. There's still the larger issue of why we have different behavior, but we can work on that later.

PR #449, opened 2023-11-21
Do not require DB fields in pdf-map

spokenbird:
[#186402592]

### DIFF
--- a/src/main/java/formflow/library/pdf/DatabaseFieldPreparer.java
+++ b/src/main/java/formflow/library/pdf/DatabaseFieldPreparer.java
@@ -17,20 +17,23 @@ public class DatabaseFieldPreparer implements DefaultSubmissionFieldPreparer {
   public Map<String, SubmissionField> prepareSubmissionFields(Submission submission, PdfMap pdfMap) {
     Map<String, SubmissionField> databaseFields = new HashMap<>();
     Map<String, Object> dbFields = pdfMap.getDbFields();
-
-    dbFields.forEach((fieldName, value) -> {
-      switch (fieldName) {
-        case "submittedAt" ->
-            databaseFields.put(fieldName, new DatabaseField(fieldName, formatDateWithNoTime(submission.getSubmittedAt())));
-        case "submissionId" -> databaseFields.put(fieldName, new DatabaseField(fieldName, submission.getId().toString()));
-        case "createdAt" ->
-            databaseFields.put(fieldName, new DatabaseField(fieldName, formatDateWithNoTime(submission.getCreatedAt())));
-        case "updatedAt" ->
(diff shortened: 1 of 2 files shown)


src/main/java/formflow/library/pdf/DatabaseFieldPreparer.java
```diff
@@ -17,20 +17,23 @@ public class DatabaseFieldPreparer implements DefaultSubmissionFieldPreparer {
...
-        case "flow" -> databaseFields.put(fieldName, new DatabaseField(fieldName, submission.getFlow()));
-        default -> log.error("Unable to map unknown database field: {}", fieldName);
-      }
-    });
+    
+    if (dbFields != null) {
+      dbFields.forEach((fieldName, value) -> {
+        switch (fieldName) {
+          case "submittedAt" ->
+              databaseFields.put(fieldName, new DatabaseField(fieldName, formatDateWithNoTime(submission.getSubmittedAt())));
```
vrajmohan, line 25: I realize that my comment has nothing to do with _this_ story. Looking around, it feels awkward that the returned `Map` duplicates the `fieldName` in the key and in the `name` property of the value.

src/main/java/formflow/library/pdf/DatabaseFieldPreparer.java
```diff
@@ -17,20 +17,23 @@ public class DatabaseFieldPreparer implements DefaultSubmissionFieldPreparer {
...
+          case "submissionId" -> databaseFields.put(fieldName, new DatabaseField(fieldName, submission.getId().toString()));
+          case "createdAt" ->
+              databaseFields.put(fieldName, new DatabaseField(fieldName, formatDateWithNoTime(submission.getCreatedAt())));
+          case "updatedAt" ->
+              databaseFields.put(fieldName, new DatabaseField(fieldName, formatDateWithNoTime(submission.getUpdatedAt())));
+          case "flow" -> databaseFields.put(fieldName, new DatabaseField(fieldName, submission.getFlow()));
+          default -> log.error("Unable to map unknown database field: {}", fieldName);
```
vrajmohan, line 32: Do we need this? Do we not "fail early" if there are spurious fields in the YAML?

vrajmohan (commented):
Minor comments

src/main/java/formflow/library/pdf/DatabaseFieldPreparer.java
spokenbird: For this case we do not. We do for the `flows-config` yaml but not for the `pdf-map` yaml. You make a good point though and we probably should be doing that. Maybe we make a ticket to do that outside of this ticket?

PR #454, opened 2023-11-27
Switch to openpdf 186505365

vrajmohan:


src/test/java/formflow/library/pdf/PDFFormFillerTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  void shouldAccept_a_with_macron() throws IOException {
+    String textFieldValue = "Michaelā";
+    Collection<PdfField> fields = List.of(
+            new PdfField("TEXT_FIELD", textFieldValue)
+    );
+
+    File file = pdfFormFiller.fill(pdf, fields, true);
+
+    assertTrue(getText(file).contains("Michael ā")); // openpdf's text extraction introduces a space
```
bseeger, line 58: That's odd. Does it do this with all diacritics? 

In general, I think it'd be great to test a list of diacritics here- mainly the ones used in Spanish and some of the Vietnamese ones.  I'd throw in an umlaut as well.

bseeger:
Do you have a good sense of what the LGPL and MPL would obligate us to? Does the application have to be open source then, or if a gov't partner wants it closed source, can they do that using a library with this license?

src/test/java/formflow/library/pdf/PDFFormFillerTest.java
vrajmohan: It is just a result of imperfect text extraction from PDF files. The PDF displays "Michaelā" correctly (no embedded space). It is not limited to odd characters - it happens for all kinds of text, even regular English words.

vrajmohan:
You can see it here - https://itsfoss.com/open-source-licenses-explained/.

src/test/java/formflow/library/pdf/PDFFormFillerTest.java
coltborg: Confirmed @bseeger, I paired with Vraj on this and the space isn't caused by the special character, spaces can happen anywhere and is more the result of `getText()`. It's not a perfect mirror of how the PDF looks.

bseeger: Is it hard to add more diacritics into this?   Just a quick run through of the more common ones?

src/main/java/formflow/library/FileController.java
```diff
@@ -36,17 +25,20 @@
...
+import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.reactive.function.client.WebClientResponseException;
 import org.springframework.web.server.ResponseStatusException;
 import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody;
 import org.springframework.web.servlet.view.RedirectView;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.*;
```
bseeger, line 37: We generally try to avoid `.*` -- can you expand this out?   We've had to adjust a setting in intellij as well to stop it from automatically doing this.

bseeger: There is one on line 28 as well.

src/test/java/formflow/library/pdf/PDFFormFillerTest.java
coltborg: We can add more diacritics, just being clear that the diacritics is not the cause of the weird space.

src/test/resources/pdf-map.yaml
```diff
@@ -1,10 +1,10 @@
 flow: ubi
-pdf: pdfs/ubi/Multipage-UBI-Form.pdf
+pdf: pdfs/Multipage-UBI-Form.pdf
```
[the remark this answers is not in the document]
coltborg: I moved it to `pdfs/testflow`.

src/main/java/formflow/library/FileController.java
coltborg: I went through and converted all `.*` imports in the repo.

src/test/java/formflow/library/pdf/PDFFormFillerTest.java
coltborg: I added another [test for more diacritics](https://github.com/codeforamerica/form-flow/pull/454/files#diff-b2c6d5ebf543755e0887ef2cda8797a3f291cfafbc5c4a0ee9f54fed7936d040R58)

src/main/java/formflow/library/FileController.java
vrajmohan: I'll fix it. 

Wanted to point out that the pattern exists in the codebase - https://github.com/codeforamerica/form-flow/blob/74d4a62f4fe206cd6654fe393a7c3ecf9599eff2/src/test/java/formflow/library/inputs/TestFlow.java#L4 and https://github.com/codeforamerica/form-flow/blob/74d4a62f4fe206cd6654fe393a7c3ecf9599eff2/src/test/java/formflow/library/framework/ConditionalNavigationTest.java#L10. The IntelliJ setting might not be working perfectly. I'll fix it in these places as well.

vrajmohan: @coltborg beat me to it !!

spokenbird (approved):

src/main/java/formflow/library/FileController.java
bseeger: Ideally we wouldn't do that.  Those snuck past.   :)   Sometimes I don't complain about test code as it won't be part of the final jar. But I can start. ;)

src/test/java/formflow/library/pdf/PDFFormFillerTest.java
```diff
@@ -0,0 +1,114 @@
...
+  void shouldAcceptOtherDiacritic() throws IOException {
+    String textFieldValue = "áàäéèêësubstringíîïóôöúûüç";
+    Collection<PdfField> fields = List.of(
+        new PdfField("TEXT_FIELD", textFieldValue)
+    );
+
+    File file = pdfFormFiller.fill(pdf, fields, true);
+
+    assertTrue(getText(file).contains("áàäéèêësubstringíîïóôöúûüç"));
+  }
```
bseeger, line 67: lol, nice!

PR #467, opened 2023-12-08
Remove TODOs 186571893

vrajmohan:


src/test/java/formflow/library/utilities/TestUtils.java
```diff
@@ -1,19 +1,10 @@
 package formflow.library.utilities;
 
-import formflow.library.data.Submission;
 import java.io.File;
 import java.net.URL;
-import java.util.HashMap;
 
 public class TestUtils {
 
-  public static void resetSubmission() {
```
coltborg, line 10: Aren't we still using this [function in cleanup](https://github.com/codeforamerica/form-flow/blob/ec70d87a58e35c6ae3b8bdc4ff91e77bdc71e085/src/test/java/formflow/library/utilities/AbstractMockMvcTest.java#L93-L96)?

src/main/java/formflow/library/ScreenController.java
```diff
@@ -3,30 +3,14 @@
 import com.smartystreets.api.exceptions.SmartyException;
 import formflow.library.address_validation.AddressValidationService;
 import formflow.library.address_validation.ValidatedAddress;
-import formflow.library.config.ActionManager;
```
coltborg, line 6: My understand was we were avoiding `.*` imports. Was this the result of a code style converting these imports? I thought our code style enforced single imports:

2023-12-14 at 18 52 13@2x

src/test/java/formflow/library/utilities/TestUtils.java
vrajmohan: It is being removed in the same commit.

coltborg (changes requested):
Left some comments, the import changes are non-blocking. The only thing I want to make sure before merging is on if `resetSubmission` is still being used or not.

src/test/java/formflow/library/utilities/TestUtils.java
coltborg: ahh I missed it, ty.

src/main/java/formflow/library/ScreenController.java
coltborg: ^ this is a nitpick and is not needed to be resolved to complete this work.

coltborg (approved):
Comment addressed

PR #469, opened 2023-12-12
only check pattern if value is not null

analoo:
Based on conversation with @bseeger @sree-cfa @spokenbird @lkemperman-cfa 

the missing null check is causing a null pointer exception in the LA doc Uploader test. 

```
Caused by:
        java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
            at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1769)
            at java.base/java.util.regex.Matcher.reset(Matcher.java:415)
            at java.base/java.util.regex.Matcher.<init>(Matcher.java:252)
            at java.base/java.util.regex.Pattern.matcher(Pattern.java:1134)
            at java.base/java.util.regex.Pattern.matches(Pattern.java:1175)
            at formflow.library.data.validators.MoneyValidator.isValid(MoneyValidator.java:15)
            at formflow.library.data.validators.MoneyValidator.isValid(MoneyValidator.java:11)```

### DIFF
--- a/src/main/java/formflow/library/data/validators/MoneyValidator.java
+++ b/src/main/java/formflow/library/data/validators/MoneyValidator.java
@@ -22,6 +22,9 @@ public class MoneyValidator implements ConstraintValidator<Money, String> {
    */
   @Override
   public boolean isValid(String value, ConstraintValidatorContext context) {
-    return value == null || Pattern.matches("(^(0|([1-9]\\d*))?(\\.\\d{1,2})?$)?", value);
+    if (value != null){
+      return Pattern.matches("(^(0|([1-9]\\d*))?(\\.\\d{1,2})?$)?", value);
+    }
+    return true;
   }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/formflow/library/data/validators/MoneyValidator.java [hunk 1]
```diff
@@ -22,6 +22,6 @@ public class MoneyValidator implements ConstraintValidator<Money, String> {
    */
   @Override
   public boolean isValid(String value, ConstraintValidatorContext context) {
-    return Pattern.matches("(^(0|([1-9]\\d*))?(\\.\\d{1,2})?$)?", value);
+    return value == null || Pattern.matches("(^(0|([1-9]\\d*))?(\\.\\d{1,2})?$)?", value);
```
bseeger, line 25: Hi Ana, while this is totally functional, it's a little hard to read easily. What do you think of doing this instead? 

```
  if (value != null) {
      return Pattern.matches("(^(0|([1-9]\\d*))?(\\.\\d{1,2})?$)?", value);
  }
  return true;
```

sree-cfa: Does this also return true if value is ""? If not, we might want to add that as well?

sree-cfa (approved):
lgtm

PR #476, opened 2023-12-22
Remove @Transactional annotation from services

vrajmohan:
It adds no benefit given the simple ways that objects are persisted in the library. It simplifies repository code - there is no need to re-read saved objects.


bseeger:
@vrajmohan should we keep pushing this forward, or do you think we should close this?  Thanks!

vrajmohan:
See https://www.pivotaltracker.com/story/show/186677658. I'll close the PR for now.

PR #484, opened 2024-01-11
Update javadocs form flow controller

enyia21:


src/main/java/formflow/library/FormFlowController.java
```diff
@@ -19,12 +19,16 @@
 @Slf4j
 public abstract class FormFlowController {
 
+  /**
+   * This is an abstract controller class used to pass services and helper methods to controllers.
```
bseeger, line 23: This is probably better described as the parent class to the other controllers.  They inherit from this and the have access to all these variables and methods.

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -41,10 +45,23 @@ public abstract class FormFlowController {
     this.messageSource = messageSource;
   }
 
+  /**
+   * Saves a {@link Submission}
+   *
+   * @param submission submission saved through the submission repository service.
+   * @return A decrypted submission
```
bseeger, line 52: I don't think you needed to say "decrypted" here. 

Instead, maybe: "The saved submission object"

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -41,10 +45,23 @@ public abstract class FormFlowController {
...
   protected Submission saveToRepository(Submission submission) {
     return saveToRepository(submission, null);
   }
 
+  /**
+   * Saves a {@link Submission}
+   *
+   * @param submission  - a submission saved through the submission repository service
+   * @param subflowName - String of a subflow name or null
+   * @return A decrypted submission
```
bseeger, line 63: Same comment as above

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -64,12 +88,27 @@ protected FlowConfiguration getFlowConfigurationByName(String flow) {
     return flowConfigurationList.get(0);
   }
 
+  /**
+   * Checks if there are any flows matching the String parameter.
+   *
+   * @param flow {@link String} of a flow name.
+   * @return {@link Boolean} True if the flow is found in the FlowConfiguration.  False if the flow is {}not found in the flowConfiguration.
+   */
   protected Boolean doesFlowExist(String flow) {
```
bseeger, line 97: The @return has a `is {}not found` which seems odd. Also, I think you can capitalize the `f` in `flowConfiguration` on this same line.

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -78,11 +117,12 @@ protected static void throwNotFoundError(String flow, String screen, String mess
...
 
   /**
-   * If the submission information exists in the HttpSession, find it in the db. If a submission is not found, create a new one.
+   * If the submission information exists in the {@link HttpSession}, find it in the db. If a submission is not found, create a new one.
    *
-   * @param httpSession The HttpSession to look in for information
+   * @param httpSession The {@link HttpSession}to look in for information
    * @param flow        The name of the flow to retrieve data about
-   * @return Submission A Submission object from the database or a new one if one was not found
+   * @exception ResponseStatusException Thrown when a {@code submission} is not found when {@code getSubmissionFromSession()} is called.
```
bseeger, line 124: This technically doesn't throw this exception. It catches it and promptly ignores it, but it doesn't throw it.   Folks reading java docs only care about things that ripple to outside of the method.

src/main/java/formflow/library/FormFlowController.java
enyia21: done

enyia21: done

src/main/java/formflow/library/FormFlowController.java
enyia21: done

src/main/java/formflow/library/FormFlowController.java
enyia21: done

src/main/java/formflow/library/FormFlowController.java
enyia21: done

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -41,10 +45,23 @@ public abstract class FormFlowController {
...
+   * @return The saved {@link Submission} object
+   */
   protected Submission saveToRepository(Submission submission) {
     return saveToRepository(submission, null);
   }
 
+  /**
+   * Saves a {@link Submission}
+   *
+   * @param submission  - a submission saved through the submission repository service
```
bseeger, line 61: I just noticed that this javadoc set has `-` after the param name, which isn't necessary.  It results in weird rendering. 
Screenshot 2024-01-12 at 9 38 00 AM

src/main/java/formflow/library/FormFlowController.java
```diff
@@ -64,12 +88,27 @@ protected FlowConfiguration getFlowConfigurationByName(String flow) {
     return flowConfigurationList.get(0);
   }
 
+  /**
+   * Checks if there are any flows matching the String parameter.
```
bseeger, line 92: I think this should be "Checks if there are any flows matching the flow name passed in."  That's more specific and easier to read.

src/main/java/formflow/library/FormFlowController.java
bseeger: The `<br>` makes this look really nice!

bseeger (approved):
Looks good!  Feel free to merge after the two new small comments are addressed.

PR #487, opened 2024-01-18
Fix error where any page with the word "submit" in its name would cause the submission to submit

bseeger:
I just discovered that if a page has the word "submit" in the name of the page, it will cause the submission to get submitted. 

Thankfully the only examples of where this happens in state projects is where we do want the data marked as submitted. 

Tests are updated as well.

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -260,6 +260,14 @@ public class ScreenController extends FormFlowController {
     // Address validation
     handleAddressValidation(submission, formSubmission);
 
+    // if there's already a session
+    if (submission.getId() != null) {
+      submission.mergeFormDataWithSubmissionData(formSubmission);
+    } else {
+      submission.setFlow(flow);
+      submission.setInputData(formSubmission.getFormData());
+    }
+
     // handle marking the record as submitted, if necessary
     if (submitSubmission) {
       log.info(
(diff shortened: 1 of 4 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -197,7 +197,7 @@ ModelAndView postScreen(
     handleAddressValidation(submission, formSubmission);
 
     // handle submit actions, if requested
-    if (request.getRequestURI().toLowerCase().contains("submit")) {
+    if (request.getRequestURI().toLowerCase().endsWith("/submit")) {
```
bseeger, line 200: This still doesn't feel good enough, but is better than before.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -196,8 +260,8 @@ ModelAndView postScreen(
     // Address validation
     handleAddressValidation(submission, formSubmission);
 
-    // handle submit actions, if requested
-    if (request.getRequestURI().toLowerCase().contains("submit")) {
+    // handle marking the record as submitted, if necessary
```
spokenbird, line 263: Maybe do this as the last thing before right before save after the data merge from formData? Thoughts?

vrajmohan (approved):
LGTM

src/main/java/formflow/library/ScreenController.java
bseeger: I moved this down past the merge but before the action. I think it's still good to have this before the action.

PR #490, opened 2024-01-22
Pull the submission id from session correctly in the MDC Filter Log

bseeger:
This will fix how the submission id is pull from the session in the MDC Insertion Filter Code.  The code needs to pull it from a session map, or better yet, use the logic in the FormFlowController parent class to get the submission id.  This code now does this.

This also changes the indent of this code to 4 spaces.


vrajmohan (approved):
Looks good to me. Why did it not work in the `mdpilot` code? Were the filters interfering with one another?

PR #494, opened 2024-01-26
Fix log message

guiburato:


bseeger:
This makes a log message function better - it was using outdated logic to get the submission information, so this PR simply moves up where the submission is retrieved and uses the id from that instead.

bseeger:
@guiburato because you and I paired on this and all the tests are passing up here, you can go ahead and merge this.

PR #495, opened 2024-01-26
Adding known translations

sree-cfa:
From JR (translation service)


coltborg (approved):
Worked as I expected locally with these starter app settings:

```yaml
form-flow:
  languages: en, es, vi
```

2024-02-08 at 13 01 52@2x

PR #500, opened 2024-02-01
Adding "condition" logic at the screen level

sree-cfa:
If client attempts to view a screen that does not meet the condition, they will be redirected to the next viewable screen.

See [notion page](https://www.notion.so/cfa/FFB-Adding-condition-at-screen-level-4e68fa7271ae4320b7fbf1e22c8c6784?pvs=4) for details

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ More examples of conditions can be found in our
 [starter application](https://github.com/codeforamerica/form-flow-starter-app/tree/main/src/main/java/org/formflowstartertemplate/app/submission/conditions)
 .
 
-### Using Conditions in `flow-config.yaml`
+### Using Conditions in `flows-config.yaml`
 
 Conditions can be configured at the screen level as a pre-condition or at the next screen level as a post-condition.
 Depending on the scenario, it may be better to configure it in one way vs the other. In either case, the application
--- a/README.md
+++ b/README.md
@@ -364,6 +364,38 @@ More examples of conditions can be found in our
 [starter application](https://github.com/codeforamerica/form-flow-starter-app/tree/main/src/main/java/org/formflowstartertemplate/app/submission/conditions)
 .
 
+### Using Conditions in `flow-config.yaml`
+
+Conditions can be configured at the screen level as a pre-condition or at the next screen level as a post-condition.
+Depending on the scenario, it may be better to configure it in one way vs the other. In either case, the application
+flow will behave the same way. The difference is that pre-conditions prevent the page from being viewed at all - the
+ScreenController will look for the next viewable page.
+
+The following 2 examples of condition configuration result in the same flow behavior.
+
+``` yaml
+flow: ubi
+  incomeQuestion:
+    nextScreens:
+    - name: incomeDetails
+  incomeDetails:
+    condition: HasIncome # pre-condition
+    nextScreens:
+      - name: expensesQuestion
+```
+
+``` yaml
+flow: ubi
+  incomeQuestion:
+    nextScreens:
+    - name: incomeDetails
+      condition: HasIncome # post-condition
+    - name: expensesQuestion
+  incomeDetails:
+    nextScreens:
+      - name: expensesQuestion
+```
+
 ### Using Conditions in Templates
 
 We have created a Java object named `ConditionManager` that's part of the model data handed to the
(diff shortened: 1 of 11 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -116,6 +116,11 @@ ModelAndView getScreen(
       return new ModelAndView("redirect:" + lockedSubmissionRedirectUrl);
     }
 
+    String redirectPath = getRedirectPath(flow, uuid, currentScreen, submission);
```
bseeger, line 119: I think this method name needs to be clearer -- it's not just getting a redirect path, it's checking if we have a condition on the page and run it if we do, returning the URL it thinks we should be on (running through all page level conditions in between to allow it to skip ahead to the next "runnable" page). 

What's a good name for that?   checkPageLevelCondition()... I don't have any good ideas that would reflect what it's doing and that it's returning a redirect Path (or not).   Is this one method doing too much?   Maybe you could do the check for the condition existing here and then call this getRedirectPath() since then it *will* be returning a path. 

Is this condition existing to aid in your development process, or will it run in live code as well?

README.md
```diff
@@ -356,6 +356,38 @@ More examples of conditions can be found in our
 [starter application](https://github.com/codeforamerica/form-flow-starter-app/tree/main/src/main/java/org/formflowstartertemplate/app/submission/conditions)
 .
 
+### Using Conditions in `flow-config.yaml`
```
bseeger, line 359: nit: it's `flows-config.yaml`

src/main/java/formflow/library/ScreenController.java
bseeger: This is really great - thanks for creating this.

sree-cfa: Yeah - will come up with a better name. 

Yeah, it's a development process aid - we want to "disable" pages that have gone through acceptance, but maybe aren't staging/production ready.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -146,6 +155,22 @@ ModelAndView getScreen(
...
+   * @param uuid          The uuid of a subflow entry
+   * @param currentScreen The current screen to check
+   * @param submission    submission
+   * @return True - current screen does not meet the condition; False - otherwise
+   */
+  private boolean shouldRedirectToNextScreen(String uuid, ScreenNavigationConfiguration currentScreen, Submission submission) {
+    if (conditionManager.conditionExists(currentScreen.getCondition())) {
+      boolean skipSubflowScreen = currentScreen.getSubflow() != null && !conditionManager.runCondition(currentScreen.getCondition(), submission, uuid);
+      return skipSubflowScreen || !conditionManager.runCondition(currentScreen.getCondition(), submission);
+    }
```
bseeger, line 170: I think I'm a little confused by this code here.   Maybe the check for if we are in a subflow could be done first here?  

For example, if we are in a subflow and the condition returns false, then it doesn't make sense to try to run the condition again as if it's not in a subflow.

bseeger:
```
 private boolean shouldRedirectToNextScreen(String uuid, ScreenNavigationConfiguration currentScreen, Submission submission) {
    if (conditionManager.conditionExists(currentScreen.getCondition())) {
      if (currentScreen.getSubflow() != null) { 
         return !conditionManager.runCondition(currentScreen.getCondition(), submission, uuid)
      }
      return !conditionManager.runCondition(currentScreen.getCondition(), submission);
    }
```

(Outside of this ticket, this makes me realize we could use a currentScreen.isInSublfow() or something like that)

src/main/java/formflow/library/ScreenController.java
bseeger: I like the re-org, I think it makes it a little clearer what's going on.

src/main/java/formflow/library/ScreenController.java
sree-cfa: Yeah, that second comment is a lot easier to read

coltborg (changes requested):
I agree with @bseeger's comment about the conditional styling.

I also think we should add `condition` to the screen in the `intellij-settings/flows-config-schema.json` so that we get typeahead support, like we get when we right a post condition in the `nextScreens` section.

coltborg:
Oh and another test case to try is to mix a `pre condition` and a `post condition` and make sure it works as expected.

coltborg (approved):
Comments have been addressed. Lgtm;

PR #504, opened 2024-02-02
Small README updates

spokenbird:


README.md
```diff
@@ -1780,11 +1780,10 @@ inputs to PDF fields.
 
 #### application.yaml properties
 
-| Name                            | Description                                                                                                                        | Example                                 |
-|---------------------------------|------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------|
-| `pdf`                           | The path to the pdf template file the FFB library should use when generating your PDF. The path should begin with a forward slash. | `pdf: /pdfs/ubi/Multipage-UBI-Form.pdf` |
```
sree-cfa, line 1785: oh, this should be "path", not pdf.

README.md
```diff
@@ -1780,11 +1780,10 @@ inputs to PDF fields.
...
-| `generate-flattened` (optional) | Boolean flag for whether to generate a flattened pdf (`true`) or not (`false`). Defaults to `true`.                                | `generate-flattened: false`             |
+| Name                            | Description                                                                                                                                                                                                                    | Example                                 |
+|---------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------|
+| `map-file`                      | The name of the yaml file where your pdf mappings live. This is `pdf-map.yaml` by default.                                                                                                                                     | `map-file: pdf-map.yaml`                |
```
bseeger, line 1785: I think the only thing I might suggest here is showing an example of this so folks see that these config params go under `pdf:` in the file itself. If I were just to read this I wouldn't know that these are not top level params.

bseeger: Sort of out of the scope of this change, but it would be handy to say for the `map-file` something like: `The server will look for this file, by default, in the resources directory.` Not sure that's the best wording, but something like it, so folks know that they are filling in the file name and where they should put the file in the first place.

bseeger (approved):
Minor comments. Looks good overall.

README.md
spokenbird: We actually realized it's not needed at all anymore after a change we made a while back 😅

README.md
spokenbird: Fixed

PR #508, opened 2024-02-10
Improve radio fieldset

guiburato:
- implement new optional param(radioHelpText) for radio input

### DIFF
--- a/src/main/resources/templates/fragments/inputs/radio.html
+++ b/src/main/resources/templates/fragments/inputs/radio.html
@@ -1,5 +1,6 @@
 <th:block
     th:fragment="radio"
+    th:with="hasHelpText=${!#strings.isEmpty(radioHelpText)}"
     th:assert="
       ${!#strings.isEmpty(inputName)},
       ${!#strings.isEmpty(value)},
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/fragments/inputs/radio.html
```diff
@@ -19,5 +20,9 @@
             aria-invalid=${hasError},
             data-follow-up=${followUpId}">
     <span th:text="${label}"></span>
+    <p th:if="${hasHelpText}"
+       th:id="${inputName} + '-' + ${value} + '-help-text'"
+       th:utext="${radioHelpText}"
+       class="text--help" style="margin-bottom: 0px"></p>
```
sree-cfa, line 26: I think this could be `class="text--help with-no-padding"`

sree-cfa, line 26:
```suggestion
       class="text--help with-no-padding"></p>
```

guiburato: Nice! Thanks.

bseeger:
Out of curiosity - the `radioFieldset` does have helptext - do you have a use case where you're not using a field set?  Or a use case where there is helptext for each radio button?

sree-cfa:
@bseeger  checkboxInSet has this at the selection-level too. We want some additional text for a radio option
Screen Shot 2024-02-07 at 3 03 48 PM

bseeger (approved):
LGTM!

PR #518, opened 2024-02-20
Adding param to screens fragments 🍖 

sree-cfa:
#### Description of change ✍️

Adding iconFragment to screens since there might be multiple icons files in other SNE apps

#### Priority 🥇

Medium - we can override these locally, but they'd be nice to have in the FFL

No breaking changes - this is adding a new param to a fragment, if it's null, it will not show up

*Didn't update docs/tests/readme since `iconName` doesn't explicitly have these either

- [x] Does the new code follow our preferred coding style?

### DIFF
--- a/intellij-settings/LiveTemplates.xml
+++ b/intellij-settings/LiveTemplates.xml
@@ -241,8 +241,9 @@
   <option name="HTML" value="true" />
 </context>
 </template>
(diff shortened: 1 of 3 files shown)


sree-cfa:
TODO - update livetemplate

coltborg (approved):
Reviewed both fragments and they act the same. Also tested the updated live templates. Lgtm; :shipit:

PR #522, opened 2024-02-24
Update fragments

sree-cfa:
#### Description of change ✍️

Updated the checkbox and radio fieldset fragments to be able to take a collection of options. 

The `options` param are values from classes that implement `InputOption`. They must have a value and label/helpText are optional. This is to simplify rendering for static checkboxes and have a more centralized way of referencing these values in html rendering and PDF generation.
*Backwards compatible.

#### Priority 🥇
Medium - we can override locally for now

No breaking changes - this will be backwards compatible

#### Testing
There's a unit test showing how to use this new param. Here's an example as well (GenderOptions will have to be implements)

```html
<th:block th:with="genderOptions=${T(org.ilgcc.app.utils.GenderOption).values()},
  <th:block th:replace="~{fragments/inputs/checkboxFieldset ::
    checkboxFieldset(inputName='childGender',
      label=#{children-ccap-info.gender-question},
      options=${genderOptions})}"/>
```

### DIFF
--- a/src/main/resources/templates/fragments/inputs/checkboxFieldset.html
+++ b/src/main/resources/templates/fragments/inputs/checkboxFieldset.html
@@ -4,6 +4,8 @@
       hasHelpText=${!#strings.isEmpty(fieldsetHelpText)},
       hasLabel=${!#strings.isEmpty(label)},
       hasAriaLabel=${!#strings.isEmpty(ariaLabel)},
+      hasOptions=${options != null},
+      hasContent=${!hasOptions},
       hasError=${
         errorMessages != null &&
         errorMessages.get(inputName) != null &&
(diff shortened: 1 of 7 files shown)


src/main/resources/templates/fragments/inputs/checkboxFieldset.html
```diff
@@ -27,7 +30,22 @@
...
-      <th:block th:replace="${content}"/>
+      <th:block th:if="${hasContent}">
+        <th:block th:replace="${content}"/>
+      </th:block>
+      <th:block th:if="${hasOptions}">
+        <th:block th:each="option : ${options}">
+          <th:block th:replace="~{fragments/inputs/checkboxInSet ::
+            checkboxInSet(inputName=${inputName},
+              value=${option.getValue()},
+              label=${#strings.isEmpty(option.getLabel()) ? null : #messages.msg(option.getLabel())},
```
sree-cfa, line 41: Should these be interpolated beforehand?

PR #527, opened 2024-03-05
Fix date errors for component

sree-cfa:
#### Description of change ✍️
- We want "form-group--error" if there is also an overall error for the date field
- cleanup - arrays.isEmpty should also check for null

#### Priority 🥇
Medium - we can override the component in the meantime

- TODO verify that this is backwards compatible

- [x] Does the code include javadocs, where necessary?
- [x] Have tests for this feature been added / updated?


enyia21 (changes requested):
Should be an explicit test for the date.html fragment.  We should test both the error messages and that the `form-group--error` class is added appropriately to the group

PR #529, opened 2024-03-06
Honeycrisp updates (Breaking)

sree-cfa:
#### Description of change ✍️
Fixing "none of the above" checkboxes to be unique so that they don't affect the function of other check box sets on the same page
See https://github.com/codeforamerica/honeycrisp-gem/pull/320

Reveal component accessibility
See https://github.com/codeforamerica/honeycrisp-gem/pull/317

#### Priority 🥇
High - We're currently running into an issue w checkboxes in il-gcc and can't easily address it in the app itself.

Possibly breaking 
- this affects checkboxes with a "none of the above" option. If the checkbox fragment is overridden, then there may need to be an update to the id for "none__checkbox" to include the inputName --> "none__checkbox-inputName"
- reveal component class changed from "reveal__link" to "reveal__button" - should be updated in SNE css classes as well

- [x] Have tests for this feature been added / updated?


spokenbird (approved):
Looks good, thanks @sree-cfa !

sree-cfa:
Note: Associated honeycrisp sha https://github.com/codeforamerica/honeycrisp-gem/commit/4d31037c3a25d5887502bd58255bafbe8108bddb

PR #547, opened 2024-04-23
Attempts to fix security errors around using unsanitized `flow` and `screen` inputs

bseeger:
#### Description of change ✍️

Creates a way to use `flow` and `screen` for redirects in a way that they are first validated to be safe values to use. 
The `getScreenConfig` was already ensuring they were good values, but that doesn't stop the code scanning for thinking they are unsafe.  

- [x] Does the code include javadocs, where necessary?
- [N/A] Have tests for this feature been added / updated?
- [N/A] Has the readme been updated?

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -19,6 +19,7 @@ import formflow.library.data.UserFileRepositoryService;
 import formflow.library.file.FileValidationService;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpSession;
+
 import java.io.IOException;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
@@ -29,6 +30,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.UUID;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.jetbrains.annotations.Nullable;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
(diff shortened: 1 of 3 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -160,6 +176,11 @@ ModelAndView getScreen(
    * Checks if current screen condition is met.
    *
    * @param uuid          The uuid of a subflow entry
+   *                      }
+   *                      <p>
+   *                      /**
+   *                      Checks if current screen condition is met.
```
bseeger, line 182: Hmmm... typo, I think. I will revert this change.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -86,29 +90,41 @@ public ScreenController(
     log.info("Screen Controller Created!");
   }
 
+  @AllArgsConstructor
+  @Getter
+  private class ScreenConfig {
+    String flowName;
+    String screenName;
+    ScreenNavigationConfiguration screenNavigationConfiguration;
+  }
```
bseeger, line 99: If we like this, I can put it out into another class. The problem I'm trying to solve is that you can't get the flow or the screen name from the `screenNavigationConfiguration` object, so I thought I'd do a light weight wrapping of it with that data, for convenience.  If you can think of a better way, let me know. 

It's hard to change the `ScreenNavigationConfiguration` object because it's tightly coupled to the yaml format.

bseeger:
This doesn't appear to fix the issue, though I'm not particularly worried about these pieces of code being a security vulnerability, given the code path.

bseeger:
I think the last change fixes the security alerts. Question is, do we like this change?

src/main/java/formflow/library/config/FlowConfiguration.java
```diff
@@ -29,4 +32,13 @@ public class FlowConfiguration {
   public ScreenNavigationConfiguration getScreenNavigation(String screenName) {
     return flow.get(screenName);
   }
+
+  public void setFlow(Map<String, ScreenNavigationConfiguration> screenMap) {
```
bseeger, line 36: This method will get called by SnakeYAML when it builds the flow configuration from the yaml file. This allows us to inject information, if we want.

coltborg (approved):
Yep, I think having the warning go away means you've addressed them 👍🏻

PR #548, opened 2024-04-25
Add validation for a few inputs (like flow and uuid) in a few spots.

bseeger:
#### Description of change ✍️

This adds validation of some request inputs in a few more spots.  This is to ensure that "flow" and "uuid" are not used w/o validating they are legitimate data. 

This should not affect any system, it will just make us a little more secure. 

- [ x] Does the new code follow [our preferred coding
- [x ] Does the code include javadocs, where necessary?
- [N/A] Has the readme been updated?

### DIFF
--- a/src/main/java/formflow/library/FormFlowController.java
+++ b/src/main/java/formflow/library/FormFlowController.java
@@ -6,11 +6,13 @@ import formflow.library.data.Submission;
 import formflow.library.data.SubmissionRepositoryService;
 import formflow.library.data.UserFileRepositoryService;
 import jakarta.servlet.http.HttpSession;
+
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.Map;
 import java.util.HashMap;
+
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.MessageSource;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


src/main/java/formflow/library/FormFlowController.java
```diff
@@ -72,10 +74,10 @@ protected Submission saveToRepository(Submission submission, String subflowName)
 
   /**
    * Gets the {@link FlowConfiguration} object for a given flow
-   * @throws ResponseStatusException when FlowConfigurations are not found.
    *
    * @param flow {@link String} of a flow name.
    * @return Returns a {@link FlowConfiguration} object.
+   * @throws ResponseStatusException when FlowConfigurations are not found.
```
bseeger, line 80: These are auto formatting, I'm not this picky. :)

src/test/resources/templates/yetAnotherTestFlow/subflowAddItemPage2.html
```diff
@@ -3,82 +3,82 @@
...
-    <div class="grid">
-      <main id="content" role="main" class="form-card spacing-above-35">
-        <th:block
-            th:replace="~{fragments/cardHeader :: cardHeader(header='Subflow Page', subtext='This is a page within a subflow')}"/>
-        <!-- Test that the currentSubflowItem is passed in by the controller: -->
-        <div th:text="${currentSubflowItem.firstNameSubflowPage2}"></div>
-        <div>
-          <th:block
-              th:replace="~{fragments/inputs/text ::
+    <div th:replace="~{fragments/toolbar :: toolbar}"></div>
```
bseeger, line 6: There are no actual changes in this file, other then format ones. I can revers this, but since we are moving to these new spacings anyways, I'll just leave this.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -908,4 +925,13 @@ private String getLockedSubmissionRedirectUrl(String flow, RedirectAttributes re
     return String.format("/flow/%s/%s", flow, lockedSubmissionRedirectPage);
   }
 
+  private String getValidatedIterationUuid(Submission submission, ScreenNavigationConfiguration screen, String uuidToVerify) {
+    Map<String, Object> iteration = submission.getSubflowEntryByUuid(screen.getSubflow(), uuidToVerify);
+    if (iteration == null) {
+      throwNotFoundError(submission.getFlow(), screen.getName(),
+          String.format("UUID not found in iterations for subflow '%s' in flow '%s', when navigating to '%s'",
```
spokenbird, line 932: Should we log the UUID?

bseeger: That would be useful. I'll add it in.

spokenbird (approved):

PR #551, opened 2024-05-02
Fix error seen in validating uuid. 

bseeger:
#### Description of change ✍️
This fix relates to something missed when validating uuids.  If a subflow iteration was already deleted and the user clicked "Go Back" the code didn't handle that well. This fixes how the code handles that scenario and adds a tests for that to catch regressions in the future. 

- [x ] Does the new code follow [our preferred coding
- [x ] Does the code include javadocs, where necessary?
- [x ] Have tests for this feature been added / updated?
- [na ] Has the readme been updated?


src/main/java/formflow/library/ScreenController.java
```diff
@@ -63,6 +64,7 @@ public class ScreenController extends FormFlowController {
 
   public static final String FLOW = "/flow";
   public static final String FLOW_SCREEN_PATH = "{flow}/{screen}";
+  private final String DELETE_CONFIRMATION_REDIRECT_UUID = "deleteConfirmationRedirectUuid";
```
bseeger, line 67: I'm going to take this out. It was for an intermediate impl attempt that we didn't end up using.

spokenbird (approved):
Looks good!

PR #575, opened 2024-07-25
Update deprecated gradle build action in github actions workflows

spokenbird:
#### Description of change ✍️
Updates the gradle build step of several actions workflows where the gradle build step was deprecated. 

See: https://github.com/marketplace/actions/gradle-build-action


coltborg (commented):
Looks inline with the [upgrades docs](https://github.com/marketplace/actions/gradle-build-action)

analoo (approved):
Thank you for linking the documentation. It makes reviewing this change super easy.

coltborg (approved):

PR #598, opened 2024-09-11
[CCAP-341] Create confirmation code for family applications

cram-cfa:
#### Issue tracking number 🔗

https://codeforamerica.atlassian.net/browse/CCAP-341


src/main/java/formflow/library/data/Submission.java
```diff
@@ -251,6 +272,60 @@ private static Map<String, Object> copyMap(Map<String, Object> origMap) {
...
+          // set this submission's shortcode, and persist it
+          shortCode = newCode;
+          submissionRepository.save(this);
+        } else {
+          log.warn("Confirmation code {} already exists", newCode);
+        }
+      } while (shortCode == null);
+    }
+    return shortCode;
+  }
```
spokenbird, line 327: Seeing this now I wonder if it makes sense to do this in the ScreenController on Post when you save the submission for the first time? 

That would avoid the need to have applicationContext and SubmissionRepository as dependencies here since the screen controller has SubmissionRepositoryService as a dependency already and you could then have ShortCodeConfig as a dependency of SubmissionRepositoryService.

That said, my Hibernate / JPA foo is not super strong, but my main concern with this is the ApplicationContext, since the library gets pulled in by other applications which will already have their own context. 

If we use this implementation it might be a good idea to locally pull this library branch into the starter-app and see if it plays nice there.

src/test/java/formflow/library/repository/SubmissionRepositoryServiceTest.java
```diff
@@ -38,6 +45,60 @@ void shouldSaveASubmissionWithUUID() {
...
+
+    Optional<Submission> reloadedSubmission = submissionRepositoryService.findByShortCode(submission.getShortCode());
+    if (reloadedSubmission.isPresent()) {
+      assertThat(submission).isEqualTo(reloadedSubmission.get());
+      assertThat(submission.getShortCode()).isEqualTo(reloadedSubmission.get().getShortCode());
+    } else {
+      Assertions.fail();
+    }
+  }
+
```
spokenbird, line 101: Maybe some tests for each of the different configurations? Annoyingly I haven't figured out a good way to load different properties per test without the need for multiple test files :(

src/main/java/formflow/library/data/Submission.java
cram-cfa: @spokenbird this is all cleaned up now in the PR. I still have to write more tests, but functionally this should be what we discussed on Tuple!

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -17,67 +19,125 @@
...
+        var subflowArr = (ArrayList<Map<String, Object>>) submission.getInputData().get(subflowName);
+
+        if (subflowArr != null) {
+            for (var entry : subflowArr) {
+                entry.remove("_csrf");
+            }
+        }
+    }
+
+    /**
```
cram-cfa, line 95: There was a bunch of reformatting above, but this is where the new method begins.

src/test/java/formflow/library/repository/SubmissionRepositoryServiceTest.java
cram-cfa: Added a few tests in new files that handle the permutations of configs. I think I should also figure out how to add this to any tests that actually hit the ScreenController too.

src/main/java/formflow/library/config/submission/ShortCodeConfig.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Value("${form-flow.short-code.creation-point:submission}")
+    private ShortCodeCreationPoint creationPoint;
+
+    @Value("${form-flow.short-code.prefix:#{null}}")
+    private String prefix;
+
+    @Value("${form-flow.short-code.suffix:#{null}}")
+    private String suffix;
+
```
cram-cfa, line 36: @spokenbird I added a prefix and suffix, defaulted to null and configurable.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -339,10 +347,19 @@ private ModelAndView handlePost(
...
       submission.setSubmittedAt(OffsetDateTime.now());
+
+      if (shortCodeConfig.isCreateShortCodeAtSubmission()) {
+        submissionRepositoryService.generateAndSetUniqueShortCode(submission);
+      }
     }
 
     actionManager.handleBeforeSaveAction(currentScreen, submission);
     submission = saveToRepository(submission);
+
```
spokenbird, line 358: I think this needs to be in the POST endpoint and not the POST + /submit endpoint.

As is right now, the at creation method would not work correctly since this endpoint is only hit when submitting the application. You want the POST mapping which starts on line 234.

spokenbird: Also, though unlikely, it's _possible_ at some point someone will create an application where the flow starts with a subflow and the first POST is in a subflow, which makes me wonder if we should account for that by adding the on creation but to the POST subflow endpoint as well (starts on 448) ?

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -17,67 +19,137 @@
...
+     * This method will check if the generated code exists in the database, and keep trying to create a unique code, before
+     * persisting and returning the newly generated code-- therefore it is very important to ensure the configuration allows for a
+     * suitably large set of possible codes for the application.
+     *
+     * @param submission the {@link formflow.library.ScreenController} for which the short code will be generated and saved
+     */
+    public synchronized void generateAndSetUniqueShortCode(Submission submission) {
+
+        if (submission.getShortCode() != null) {
+            return;
```
spokenbird, line 120: We should log here as ideally we shouldn't bit this return, but it would be to know if we are.

src/test/java/formflow/library/controllers/ScreenControllerShortCodeCreationPointTest.java
```diff
@@ -0,0 +1,96 @@
...
+    );
+
+    // Assert that the submissions submittedAt value is null before submitting
+    Map<String, UUID> submissionMap = (Map) session.getAttribute(SUBMISSION_MAP_NAME);
+    Optional<Submission> testFlowSubmission = submissionRepositoryService.findById(submissionMap.get("testFlow"));
+    assertThat(testFlowSubmission.isPresent()).isTrue();
+    assertThat(testFlowSubmission.get().getSubmittedAt()).isNull();
+//    assertThat(testFlowSubmission.get().getShortCode()).isNotNull();
+
+    ResultActions result = mockMvc.perform(post("/flow/testFlow/pageWithCustomSubmitButton/submit").session(session));
```
spokenbird, line 66: We'll need to update this test when moving the above creation method as the /submit endpoint is only meant for completing an application and setting the submittedAt, etc.

src/main/java/formflow/library/ScreenController.java
cram-cfa: Ok, these should be updated per our Tuple!

PR #613, opened 2024-10-21
[CCAP-420] Updating FFL short code configs to be set per flow and not per project

cram-cfa:
#### Issue tracking number 🔗

https://codeforamerica.atlassian.net/browse/CCAP-420

#### Description of change ✍️

This change will allow each flow to have a specific short code behavior, and if the flow is not found at all within the mapping, the short code creation will be skipped completely.

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -338,6 +338,8 @@ public class ScreenController extends FormFlowController {
       submission.setInputData(formSubmission.getFormData());
     }
 
+    ShortCodeConfig.Config config = shortCodeConfig.getConfig(flow);
+
     // handle marking the record as submitted, if necessary
     if (submitSubmission) {
       log.info(
@@ -348,7 +350,7 @@ public class ScreenController extends FormFlowController {
       );
       submission.setSubmittedAt(OffsetDateTime.now());
 
-      if (shortCodeConfig.isCreateShortCodeAtSubmission()) {
+      if (config != null && config.isCreateShortCodeAtSubmission()) {
         submissionRepositoryService.generateAndSetUniqueShortCode(submission);
       }
     }
@@ -356,7 +358,7 @@ public class ScreenController extends FormFlowController {
     actionManager.handleBeforeSaveAction(currentScreen, submission);
     submission = saveToRepository(submission);
 
-    if (shortCodeConfig.isCreateShortCodeAtCreation()) {
+    if (config != null && config.isCreateShortCodeAtCreation()) {
       submissionRepositoryService.generateAndSetUniqueShortCode(submission);
     }
 
@@ -540,7 +542,7 @@ public class ScreenController extends FormFlowController {
     actionManager.handleBeforeSaveAction(currentScreen, submission, iterationUuid);
     submission = saveToRepository(submission, subflowName);
 
-    if (shortCodeConfig.isCreateShortCodeAtCreation()) {
+    if (shortCodeConfig.getConfig(flow) != null && shortCodeConfig.getConfig(flow).isCreateShortCodeAtCreation()) {
       submissionRepositoryService.generateAndSetUniqueShortCode(submission);
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/formflow/library/config/submission/ShortCodeConfig.java
```diff
@@ -1,44 +1,56 @@
...
-        alphanumeric, alpha, numeric;
+    private Map<String, Config> shortCodeConfigs;
+
+    public Config getConfig(String flowName) {
+        return shortCodeConfigs != null ? shortCodeConfigs.get(flowName) : null;
     }
 
-    public enum ShortCodeCreationPoint {
-        creation, submission
+    public void setShortCodeConfigs(Map<String, Config> shortCodeConfigs) {
```
spokenbird, line 19: Good idea with the map

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -116,32 +116,38 @@ public void removeSubflowCSRF(Submission submission, String subflowName) {
      */
     public synchronized void generateAndSetUniqueShortCode(Submission submission) {
 
+        if (shortCodeConfig.getConfig(submission.getFlow()) == null) {
```
spokenbird, line 119: In theory this shouldn't happen since we null check in the ScreenController in the places this is called yea?

cram-cfa: Yeah, I just wanted to be sure that if someone ever used it elsewhere, we were protected against a mistake!

src/main/java/formflow/library/ScreenController.java
```diff
@@ -338,6 +338,8 @@ private ModelAndView handlePost(
       submission.setInputData(formSubmission.getFormData());
     }
 
+    ShortCodeConfig.Config config = shortCodeConfig.getConfig(flow);
```
spokenbird, line 341: Should we add a check for if this is null with a helpful log to the user? 

Maybe something like `No shortcode configuration found for flow {}, no shortcode will be generated.`

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -116,32 +116,38 @@ public void removeSubflowCSRF(Submission submission, String subflowName) {
...
+            return;
+        }
+
         if (submission.getShortCode() != null) {
             log.warn("Unable to create short code for submission {}", submission.getId());
```
spokenbird, line 124: Is this because one would already exist? If so it might help to make the log a little clearer by saying so.

src/main/java/formflow/library/data/SubmissionRepositoryService.java
```diff
@@ -116,32 +116,38 @@ public void removeSubflowCSRF(Submission submission, String subflowName) {
...
             return;
         }
 
         log.info("Attempting to create short code for submission {}", submission.getId());
 
+        ShortCodeConfig.Config config = shortCodeConfig.getConfig(submission.getFlow());
+
         // If there is no short code for this submission in the database, generate one
-        int codeLength = shortCodeConfig.getCodeLength();
+        int codeLength = config.getCodeLength();
```
spokenbird, line 133: Is it possible to get to this point with a null config?

src/main/java/formflow/library/ScreenController.java
cram-cfa: Yeah, good call here!

src/main/java/formflow/library/data/SubmissionRepositoryService.java
cram-cfa: Yep, but i agree that the warn message should just more clear.

src/main/java/formflow/library/data/SubmissionRepositoryService.java
cram-cfa: It shouldn't be! I just reorganized things slightly so the initial null check happens right before this point.

spokenbird (approved):
Looks good, thanks for adding those log changes!

PR #621, opened 2024-11-20
ILGCC-CCAP-319 Add autocomplete capabilities to relevent input fields

spokenbird:
Pretty much does what the PR title says.


src/test/java/formflow/library/file/UploadJourneyTests.java
```diff
@@ -60,7 +60,7 @@ void documentUploadFlow() {
     //Race condition caused by uploadPasswordProtectedPdf waits until upload file has file details added instead
     //of waiting until file upload is complete.
     try {
-      Thread.sleep(500);
+      Thread.sleep(2000);
```
cram-cfa, line 63: oh no, we're going to wait 2 seconds now? lol

cram-cfa (approved):

src/test/java/formflow/library/file/UploadJourneyTests.java
spokenbird: For some reason it kept faling at 500 and 1000 :/

PR #628, opened 2024-12-09
Update library to springboot 3.4.0

spokenbird:
Updates the library to use Springboot 3.4.0 which requires updating deprecated libraries and replacing some dependencies.


build.gradle
```diff
@@ -75,7 +75,7 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-webflux'
     implementation 'org.jetbrains:annotations:26.0.1'
     implementation 'io.hypersistence:hypersistence-utils-hibernate-63:3.9.0'
-    implementation 'javax.persistence:javax.persistence-api:2.2'
+    implementation 'jakarta.persistence:jakarta.persistence-api'
```
enyia21, line 78: Do we need to lock a version of `jakarta.persistence-api`?

enyia21 (approved):

cram-cfa (approved):

build.gradle
spokenbird: Good question, I actually updated to remove this explicit dependency call in favor of using the transitive one from Spring Boot.

PR #645, opened 2025-01-29
[CCAP-629][CCAP-575] Support for converting documents and images to pdfs 

cram-cfa:
#### Issue tracking number 🔗

https://codeforamerica.atlassian.net/browse/CCAP-629


src/main/java/formflow/library/FileController.java
```diff
@@ -219,6 +231,82 @@ public ResponseEntity<?> upload(
...
+                        .originalName(convertedMultipartFile.getOriginalFilename())
+                        .repositoryPath(convertedFileUploadLocation)
+                        .filesize((float) convertedMultipartFile.getSize())
+                        .mimeType(convertedMultipartFile.getContentType())
+                        .virusScanned(true)
+                        .docTypeLabel(defaultDocType)
+                        .conversionSourceFileId(userFileId)
+                        .build();
+
+                uploadedConvertedFile = userFileRepositoryService.save(uploadedConvertedFile);
```
analoo, line 297: The only part that is still unclear to me is where the conversionSourceFileId is set.

src/main/java/formflow/library/FileController.java
```diff
@@ -208,6 +218,8 @@ public ResponseEntity<?> upload(
       userFileMap.addUserFileToMap(flow, inputName, uploadedFile, thumbDataUrl);
       httpSession.setAttribute(SESSION_USERFILES_KEY, objectMapper.writeValueAsString(userFileMap));
 
+      convertUploadedFileToPDF(file, flow, inputName, userFileId, submission);
```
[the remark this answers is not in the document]
cram-cfa: There's no particular reason, I wanted the logging when I was testing things, but you're probably correct in that it'll be more consistent to move it from the service to the controller.

src/main/java/formflow/library/FileController.java
cram-cfa: This is generated on line 191; It is the UUID of the entry for the file that was uploaded by the user. It has to be generated in the code -- and not by the database -- so we can properly set the path on the uploaded file. (This was actually a bug before, but since we don't really care what the path is, as long as it's set... it's not actually causing any issues in production, etc)

Once the uploaded file has a UUID, we want the new converted file to have a reference back to that uploaded file -- hence it being the "conversion source file's id". 

The code to create the UUID previously was way up at the top of the method, I just moved it to line 191 where it is first ever needed.

Let me know if that explains it!

src/main/java/formflow/library/file/FileConversionService.java
```diff
@@ -0,0 +1,193 @@
+package formflow.library.file;
+
```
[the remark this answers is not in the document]
cram-cfa: Nope. This is what does all the work to convert files-- the changes in the IL project are just to install `soffice`, used below on line 143, into the environment itself (dev or aptible).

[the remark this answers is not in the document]
cram-cfa: The updated readme mentions installation; once I know that installation works for Aptible, I'll update that to be a bit more detailed.

src/main/java/formflow/library/file/FileConversionService.java
```diff
@@ -0,0 +1,403 @@
...
+    private final Tika tikaFileValidator;
+
+    private final FileValidationService fileValidationService;
+
+    public FileConversionService(FileValidationService fileValidationService) {
+        tikaFileValidator = new Tika();
+        this.fileValidationService = fileValidationService;
+    }
+
+    public Set<MultipartFile> convertFileToPDF(MultipartFile file) {
```
mtodd-cfa, line 81: Since this is returning a collection, could it return an empty set instead of null? It would also make it consistent with the conversion functions it calls

src/main/java/formflow/library/data/UserFileRepository.java
```diff
@@ -21,10 +21,13 @@ public interface UserFileRepository extends JpaRepository<UserFile, UUID> {
...
   /**
-   * Gets a count of all the {@link UserFile}s associated with a {@link Submission}}
+   * Gets a count of all the {@link UserFile}s associated with a {@link Submission}} where the
+   * file has never been converted
    *
    * @param submission the {@link Submission} for which the count of associated {@link UserFile}s are sought
    * @return count of {@link UserFile}s
    */
-  long countBySubmission(Submission submission);
+  long countBySubmissionAndConversionSourceFileIdIsNull(Submission submission);
```
mtodd-cfa, line 30: nit: similar to the other, this naming seems unnecessary

src/main/java/formflow/library/FileController.java
```diff
@@ -219,6 +235,93 @@ public ResponseEntity<?> upload(
...
+        // and recreate the Multipart File for conversion.
+        byte[] fileContent = java.nio.file.Files.readAllBytes(tempFile.toPath());
+        MultipartFile multipartFile = new MockMultipartFile(
+                file.getName(),
+                file.getOriginalFilename(),
+                file.getContentType(),
+                fileContent
+        );
+        return fileConversionService.convertFileToPDF(multipartFile);
+      } catch (IOException e) {
```
spokenbird, line 259: Curious why catch IO here and then any Exception in the exceptionally block?

Could we not just catch Exception here making the exceptionally redundant?

src/main/java/formflow/library/FileController.java
```diff
@@ -219,6 +235,93 @@ public ResponseEntity<?> upload(
...
+
+    // Need this to be final, for the lambda below
+    final Submission finalSubmission = submission;
+
+    fileConversion.thenAccept(convertedMultipartFiles -> {
+      // We've waited around for the original conversion call to complete and return from its thread,
+      // and now we can save and upload the file(s), if the original was converted.
+      if (convertedMultipartFiles != null && !convertedMultipartFiles.isEmpty()) {
+
+        if (convertedMultipartFiles.size() == 1) {
```
spokenbird, line 279: Why would a single file be converted into multiple PDFs?

src/main/java/formflow/library/data/UserFileRepository.java
cram-cfa: Per Slack, this needs to stay as-is so the Jpa/Hibernate code knows what to do.

src/main/java/formflow/library/FileController.java
cram-cfa: It wouldn't. but this was just leftover code from something I apparently only partially cleaned up. Should be a lot better now where we will just upload the N files returned in the Set... which will usually be 1, could be 0 if it's already a PDF, and could be 2+ if it's a word doc that needs to be split up!

src/main/java/formflow/library/FileController.java
cram-cfa: Good catch! Fixed that up.

src/main/java/formflow/library/file/FileConversionService.java
```diff
@@ -0,0 +1,403 @@
...
+        try {
+            MimeType fileMimeType = MimeType.valueOf(tikaFileValidator.detect(file.getInputStream()));
+
+            if (MIME_TYPE_MAP.get(fileMimeType) == null) {
+                log.error("Unable to convert Mime Type to PDF {}", fileMimeType);
+                return null;
```
mtodd-cfa, line 87: Last comment: can this return an empty set too? Then we don't have to worry about this method accidentally causing an NPE, even in an error state

mtodd-cfa (changes requested):

src/main/java/formflow/library/file/FileConversionService.java
cram-cfa: Oh yeah, great catch! Just another place where I missed that I was still returning `null` once I switched to a Set. I'll clean this up now.

mtodd-cfa (approved):

spokenbird (approved):

PR #658, opened 2025-02-12
[CCAP-670] Enhance file conversion service to retry up to 10 times if LibreOffice fails

cram-cfa:

### DIFF
--- a/src/main/java/formflow/library/file/FileConversionService.java
+++ b/src/main/java/formflow/library/file/FileConversionService.java
@@ -100,7 +100,7 @@ public class FileConversionService {
                 }
                 case OFFICE_DOCUMENT -> {
                     log.info("Converting document of type {} to PDF", fileMimeType);
-                    yield convertOfficeDocumentToPDF(file);
+                    yield convertOfficeDocumentToPDF(file, 0);
                 }
                 default -> {
                     log.info("Skipping converting file of type {} to PDF", fileMimeType);
@@ -152,6 +152,7 @@ public class FileConversionService {
                     log.info("Compressed file size {}", convertedPDF.getSize());
 
                     if (!isTooLarge(convertedPDF)) {
+                        log.info("Successfully converted image to PDF with file size: {}", convertedPDF.getSize());
                         break;
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/file/FileConversionService.java
```diff
@@ -274,13 +275,12 @@ private Set<MultipartFile> convertOfficeDocumentToPDF(MultipartFile file) {
...
+                if (numberOfRetries < 10) {
+                    numberOfRetries++;
+                    log.warn("Unable to find PDF file {} with exit code: {}, will retry with attempt {}", pdfFile.getAbsolutePath(), exitCode, numberOfRetries);
+                    return convertOfficeDocumentToPDF(file, numberOfRetries);
                 } else {
-                    // There is a race condition where soffice will say everything is great, but it's not great and there's no
-                    // converted file. We can retry in this situation.
-                    log.warn("Unable to find PDF file {}, will retry.", pdfFile.getAbsolutePath());
-                    return convertOfficeDocumentToPDF(file);
+                    throw new RuntimeException("PDF conversion failed. Unable to find PDF file " + pdfFile.getAbsolutePath() + " after  " + numberOfRetries + " retries. Final exit code: " + exitCode);
```
spokenbird, line 283: What happens when this exception is thrown after 10 retries?

cram-cfa: It gets all the way up to the FileController: https://github.com/codeforamerica/form-flow/blob/bd983f2a93ceacbb5fc1df91c05c3f7d11afe4a5/src/main/java/formflow/library/FileController.java#L260

And then nothing happens except logging that nothing was converted: https://github.com/codeforamerica/form-flow/blob/bd983f2a93ceacbb5fc1df91c05c3f7d11afe4a5/src/main/java/formflow/library/FileController.java#L280C19-L280C42

spokenbird: I think it bubbling up to the FileController makes sense, but maybe we should make a slack alert in Datadog if this happens so we can be alerted, as my understanding is if this happens, then it's possible we don't end up sending a supporting document with an application?

spokenbird: Ok sweet! Thanks for being thorough!

PR #667, opened 2025-02-27
[CCAP-661] - Makes string query case insensitive

analoo:


cram-cfa (approved):

cram-cfa (changes requested):
Actually, we can not do this.

It is configurable that a short_code can be all upper, all lower, or mixed cases. 

So we should not do this, and instead we should do a `.toUpperCase()` to the java code in our application for the inputted code.

cram-cfa:
Fixed without needing this, per the comments.

PR #684, opened 2025-03-31
[CCAP-766] Making createdAt and updatedAt handle being immutable at correct times

cram-cfa:


cram-cfa:
@spokenbird This seems to solve the issue in my dev. The tests pass for this, plus all our tests pass in the Illinois project when using this snapshot locally. 

The tl;dr is that we were telling Hibernate that our database would generate and set/update the timestamps, but then our db didn't have a default/trigger to do that. The easier solution is to do what this PR does, which is rely on Hibernate to appropriately set those values.

I think this is the fix, and refreshing my local DB while walking through things and saving things shows the updated_at column values updating appropriately.

PR #689, opened 2025-04-11
[CCAP-767][CCAP-828][CCAP-829] Upgrading to AWS s3 SDK v2

cram-cfa:

### DIFF
--- a/src/main/java/formflow/library/file/S3CloudFileRepository.java
+++ b/src/main/java/formflow/library/file/S3CloudFileRepository.java
@@ -1,33 +1,31 @@
 package formflow.library.file;
 
-import com.amazonaws.AmazonServiceException;
-import com.amazonaws.SdkClientException;
-import com.amazonaws.auth.AWSCredentials;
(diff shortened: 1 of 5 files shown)


src/main/java/formflow/library/file/S3CloudFileRepository.java
```diff
@@ -38,21 +36,19 @@
...
+            @Value("${form-flow.aws.s3_bucket_name}") String s3BucketName,
+            @Value("${form-flow.aws.region}") String region) {
+
         bucketName = s3BucketName;
-        s3Client = AmazonS3ClientBuilder
-                .standard()
-                .withRegion(region)
-                .withCredentials(new AWSStaticCredentialsProvider(credentials))
+
+        AwsBasicCredentials awsCreds = AwsBasicCredentials.create(accessKey, secretKey);
```
spokenbird, line 48: It's possible this might not work in QA, but I'm not sure. We may just have to make the change here and then see.

spokenbird (approved):
Looks good!

PR #738, opened 2025-05-29
[CCAP-994] Support for deleteConfirmationScreen to be null/unset

cram-cfa:
#### Issue tracking number 🔗
https://codeforamerica.atlassian.net/browse/CCAP-994

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -607,7 +607,7 @@ public class ScreenController extends FormFlowController {
     }
 
     if (submission != null) {
-      Map<Object, String> entryToDelete = (Map) submission.getSubflowEntryByUuid(subflow, uuid);
+      Map<String, Object> entryToDelete = submission.getSubflowEntryByUuid(subflow, uuid);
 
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -607,7 +607,7 @@ ModelAndView deleteConfirmation(
     }
 
     if (submission != null) {
-      Map<Object, String> entryToDelete = (Map) submission.getSubflowEntryByUuid(subflow, uuid);
+      Map<String, Object> entryToDelete = submission.getSubflowEntryByUuid(subflow, uuid);
```
cram-cfa, line 610: I am pretty sure this was a typo, since everything in our Json is a String --> Object relationship, but this has worked because a String _is_ an Object and also the subflow configuration is `key: value` like `entryScreen: something` and thus both are String... so really, in this scenario, it's `Map<String, String>` 😄 

Anyway, just figured I'd fix it.

mtodd-cfa (approved):

spokenbird (approved):
Looks good!

PR #770, opened 2025-06-16
Cleaning up some logging and other warnings

cram-cfa:

### DIFF
--- a/src/main/java/formflow/library/ScreenController.java
+++ b/src/main/java/formflow/library/ScreenController.java
@@ -183,21 +183,19 @@ public class ScreenController extends FormFlowController {
                 }
                 ;
             }
-
-            if (repeatForIterationUuid == null) {
-                // catch to see if they are trying to go to a delete confirmation screen when the UUID is not present anymore.
-                // If so, redirect.
-                if (isDeleteConfirmationScreen(flow, screen)) {
-                    ModelAndView nothingToDeleteModelAndView = handleDeleteBackBehavior(flow, screen, requestUuid, submission);
-                    if (nothingToDeleteModelAndView != null) {
-                        return nothingToDeleteModelAndView;
-                    }
-                } else {
-                    throwNotFoundError(submission.getFlow(), currentScreen.getName(),
-                            String.format(
-                                    "UUID ('%s') not found in iterations for subflow '%s' in flow '%s', when navigating to '%s'",
-                                    requestUuid, currentScreen.getSubflow(), submission.getFlow(), currentScreen.getName()));
+        } else if (repeatForIterationUuid == null) {
+            // catch to see if they are trying to go to a delete confirmation screen when the UUID is not present anymore.
+            // If so, redirect.
+            if (isDeleteConfirmationScreen(flow, screen)) {
+                ModelAndView nothingToDeleteModelAndView = handleDeleteBackBehavior(flow, screen, requestUuid, submission);
+                if (nothingToDeleteModelAndView != null) {
+                    return nothingToDeleteModelAndView;
                 }
+            } else {
+                throwNotFoundError(submission.getFlow(), currentScreen.getName(),
+                        String.format(
+                                "UUID ('%s') not found in iterations for subflow '%s' in flow '%s', when navigating to '%s'",
+                                requestUuid, currentScreen.getSubflow(), submission.getFlow(), currentScreen.getName()));
             }
         }
 
@@ -212,7 +210,7 @@ public class ScreenController extends FormFlowController {
             if (validatedUuid != null && repeatForIterationUuid != null) {
                 return new ModelAndView(String.format("redirect:/flow/%s/%s/%s/%s", flow, nextViewableScreen, validatedUuid,
                         repeatForIterationUuid));
-            } else if (validatedUuid != null && repeatForIterationUuid == null) {
+            } else if (validatedUuid != null) {
                 return new ModelAndView(String.format("redirect:/flow/%s/%s/%s", flow, nextViewableScreen, validatedUuid));
             } else {
                 return new ModelAndView(String.format("redirect:/flow/%s/%s", flow, nextViewableScreen));
@@ -231,7 +229,7 @@ public class ScreenController extends FormFlowController {
 
         if (validatedUuid != null && repeatForIterationUuid != null) {
             actionManager.handleBeforeDisplayAction(currentScreen, submission, validatedUuid, repeatForIterationUuid);
-        } else if (validatedUuid != null && repeatForIterationUuid == null) {
+        } else if (validatedUuid != null) {
             actionManager.handleBeforeDisplayAction(currentScreen, submission, validatedUuid);
         } else {
             actionManager.handleBeforeDisplayAction(currentScreen, submission);
@@ -276,7 +274,7 @@ public class ScreenController extends FormFlowController {
         if (conditionManager.conditionExists(currentScreen.getCondition())) {
             if (currentScreen.getSubflow() != null && repeatForUuid != null) {
                 return !conditionManager.runCondition(currentScreen.getCondition(), submission, uuid, repeatForUuid);
-            } else if (currentScreen.getSubflow() != null && repeatForUuid == null) {
+            } else if (currentScreen.getSubflow() != null) {
                 return !conditionManager.runCondition(currentScreen.getCondition(), submission, uuid);
             } else {
                 return !conditionManager.runCondition(currentScreen.getCondition(), submission);
@@ -375,7 +373,7 @@ public class ScreenController extends FormFlowController {
             boolean submitSubmission
     ) throws SmartyException, IOException, InterruptedException {
 
-        log.info("POST postScreen (url: {}): flow: {}, screen: {}", request.getRequestURI().toLowerCase(), requestFlow,
+        log.info("POST handlePost (url: {}): flow: {}, screen: {}", request.getRequestURI().toLowerCase(), requestFlow,
                 requestScreen);
         // Checks if screen and flow exist. If getScreenConfig runs successfully, then the flow and screen exist.
         ScreenConfig screenConfig = getValidatedScreenConfiguration(requestFlow, requestScreen);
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/ScreenController.java
```diff
@@ -231,7 +229,7 @@ ModelAndView getScreen(
 
         if (validatedUuid != null && repeatForIterationUuid != null) {
             actionManager.handleBeforeDisplayAction(currentScreen, submission, validatedUuid, repeatForIterationUuid);
-        } else if (validatedUuid != null && repeatForIterationUuid == null) {
+        } else if (validatedUuid != null) {
```
cram-cfa, line 232: IntelliJ correctly noted that `repeatForIterationUuid` can not be `null` at this point, so removing the null check here.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -276,7 +274,7 @@ private boolean shouldRedirectToNextScreen(String uuid, String repeatForUuid, Sc
         if (conditionManager.conditionExists(currentScreen.getCondition())) {
             if (currentScreen.getSubflow() != null && repeatForUuid != null) {
                 return !conditionManager.runCondition(currentScreen.getCondition(), submission, uuid, repeatForUuid);
-            } else if (currentScreen.getSubflow() != null && repeatForUuid == null) {
+            } else if (currentScreen.getSubflow() != null) {
```
cram-cfa, line 277: IntelliJ correctly noted that `repeatForUuid ` can not be `null` at this point, so removing the null check here.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -375,7 +373,7 @@ private ModelAndView handlePost(
             boolean submitSubmission
     ) throws SmartyException, IOException, InterruptedException {
 
-        log.info("POST postScreen (url: {}): flow: {}, screen: {}", request.getRequestURI().toLowerCase(), requestFlow,
+        log.info("POST handlePost (url: {}): flow: {}, screen: {}", request.getRequestURI().toLowerCase(), requestFlow,
```
cram-cfa, line 376: Wrong method name in this logging.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -539,7 +537,7 @@ RedirectView postSubflowScreen(
             Locale locale
     ) throws ResponseStatusException, SmartyException, IOException, InterruptedException {
         log.info(
-                "POST updateOrCreateIteration (url: {}): flow: {}, screen: {}, uuid: {}",
+                "POST postSubflowScreen (url: {}): flow: {}, screen: {}, uuid: {}",
```
cram-cfa, line 540: Wrong method name in this logging.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -734,11 +732,11 @@ ModelAndView deleteSubflowIteration(
             if (!subflowArr.isEmpty()) {
                 existingInputData.put(subflow, subflowArr);
                 submission.setInputData(existingInputData);
-                submission = saveToRepository(submission, subflow);
+                saveToRepository(submission, subflow);
             } else {
                 existingInputData.remove(subflow);
                 submission.setInputData(existingInputData);
-                submission = saveToRepository(submission, subflow);
+                saveToRepository(submission, subflow);
```
cram-cfa, line 739: Unused variable assignment, so I just dropped it.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -784,7 +782,7 @@ ModelAndView getRepeatForSubflowScreen(
             Locale locale
     ) throws ResponseStatusException {
         log.info(
-                "GET getSubflowScreen (url: {}): flow: {}, screen: {}, uuid: {}",
+                "GET getRepeatForSubflowScreen (url: {}): flow: {}, screen: {}, uuid: {}",
```
cram-cfa, line 785: Wrong method name in this logging.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -901,7 +899,7 @@ RedirectView postRepeatForSubflowScreen(
             Locale locale
     ) throws ResponseStatusException {
         log.info(
-                "POST updateOrCreateIteration (url: {}): flow: {}, screen: {}, uuid: {}",
+                "POST postRepeatForSubflowScreen (url: {}): flow: {}, screen: {}, uuid: {}",
```
cram-cfa, line 902: Incorrect method name in the logging.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -996,7 +994,7 @@ ModelAndView deleteRepeatForConfirmation(
             RedirectAttributes redirectAttributes,
             Locale locale
     ) {
-        log.info("GET deleteConfirmation (url: {}): flow: {}, uuid: {}", request.getRequestURI().toLowerCase(), flow, uuid);
+        log.info("GET deleteRepeatForConfirmation (url: {}): flow: {}, uuid: {}", request.getRequestURI().toLowerCase(), flow, uuid);
```
cram-cfa, line 997: Incorrect method name in the logging.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -1131,10 +1129,6 @@ ModelAndView navigation(
 
         Submission submission = getSubmissionFromSession(httpSession, flow);
         log.info("Current submission ID is: {} and current Session ID is: {}", submission.getId(), httpSession.getId());
-        if (submission == null) {
-            throwNotFoundError(flow, screen,
-                    String.format("Submission not found in session for flow '{}', when navigating to '{}'", flow, screen));
-        }
```
cram-cfa, line 1137: Can't ever have a `null` submission here, so this code can't ever happen.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -1483,8 +1477,8 @@ private Map<String, Object> createModel(String flow, String screen, HttpSession
                     if (uuidOfIterationToUpdate == null) {
                         throwNotFoundError(flow, screen,
                                 String.format(
-                                        "UUID was null when trying to find iteration for subflow '%s' in flow '%s. It's possible the user hit back in a subflow. Redirecting.'",
-                                        uuidOfIterationToUpdate, subflowName, submission.getFlow(), screen));
+                                        "UUID was null when trying to find iteration for subflow '%s' in flow '%s'. It's possible the user hit back in a subflow from screen '%s'. Redirecting.'",
+                                        subflowName, submission.getFlow(), screen));
```
cram-cfa, line 1481: This logging only had 2 variables (`%s`) but 4 inputs. Fixed it, but need confirmation that this is a good logging statement?

src/main/java/formflow/library/ScreenController.java
```diff
@@ -1660,8 +1654,8 @@ private boolean isValidRepeatForIterationUuid(Submission submission, String flow
 
         if (!subflowIterationData.containsKey(repeatForSaveDataAs)) {
             log.error(String.format(
-                    "RepeatFor SaveDataAs ('%s') key was not found in the subflow '%s' in flow '%s' with iteration ('%s'), when validating repeatFor iteration for uuid ('%s')'",
-                    repeatForSaveDataAs, subflowName, submission.getFlow()), subflowIterationUuid, repeatForIterationUuid);
+                    "RepeatFor SaveDataAs ('%s') key was not found in the subflow '%s' in flow '%s' with iteration ('%s'), when validating repeatFor iteration for uuid ('%s')",
+                    repeatForSaveDataAs, subflowName, submission.getFlow(), subflowIterationUuid, repeatForIterationUuid));
```
cram-cfa, line 1658: ) was in the wrong place, so the logging statement would have been busted.

src/main/java/formflow/library/ScreenController.java
analoo: Looks good. Thank you!

src/main/java/formflow/library/ScreenController.java
analoo: I'll defer to @spokenbird on this one. I think it is possible to have an expired session that does not have an active submission.

src/main/java/formflow/library/ScreenController.java
```diff
@@ -183,21 +183,19 @@ ModelAndView getScreen(
...
-                    ModelAndView nothingToDeleteModelAndView = handleDeleteBackBehavior(flow, screen, requestUuid, submission);
-                    if (nothingToDeleteModelAndView != null) {
-                        return nothingToDeleteModelAndView;
-                    }
-                } else {
-                    throwNotFoundError(submission.getFlow(), currentScreen.getName(),
-                            String.format(
-                                    "UUID ('%s') not found in iterations for subflow '%s' in flow '%s', when navigating to '%s'",
-                                    requestUuid, currentScreen.getSubflow(), submission.getFlow(), currentScreen.getName()));
+        } else if (repeatForIterationUuid == null) {
```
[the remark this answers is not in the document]
analoo: That makes sense. I agree with this change.

analoo (approved):

src/main/java/formflow/library/ScreenController.java
```diff
@@ -181,23 +181,6 @@ ModelAndView getScreen(
...
-                    if (nothingToDeleteModelAndView != null) {
-                        return nothingToDeleteModelAndView;
-                    }
-                } else {
-                    throwNotFoundError(submission.getFlow(), currentScreen.getName(),
-                            String.format(
-                                    "UUID ('%s') not found in iterations for subflow '%s' in flow '%s', when navigating to '%s'",
-                                    requestUuid, currentScreen.getSubflow(), submission.getFlow(), currentScreen.getName()));
-                }
             }
```
cram-cfa, line 184: Original comment still stands, that this can't ever be reached because `repeatForIterationUuid` is already checked on line 170 and confirmed to be not null!

src/main/java/formflow/library/ScreenController.java
cram-cfa: The `getSubmissionFromSession` throws an exception if a submission is never ever found, best I can tell.

PR #807, opened 2025-07-15
[CCAP-1155] Enable a secret setting to allow a read-only PDF to be not quite as read-only

cram-cfa:
#### Issue tracking number 🔗
https://codeforamerica.atlassian.net/browse/CCAP-1155

### DIFF
--- a/src/main/java/formflow/library/file/FileConversionService.java
+++ b/src/main/java/formflow/library/file/FileConversionService.java
@@ -31,6 +31,7 @@ import javax.imageio.stream.ImageOutputStream;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.tika.Tika;
+import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.MediaType;
 import org.springframework.mock.web.MockMultipartFile;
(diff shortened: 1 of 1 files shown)


src/main/java/formflow/library/file/FileConversionService.java
```diff
@@ -386,6 +395,61 @@ private Set<MultipartFile> convertOfficeDocumentToPDF(MultipartFile file, int nu
...
+
+        File tempFile = null;
+        try {
+            tempFile = File.createTempFile("upload_", "_" + file.getOriginalFilename());
+
+            try (FileOutputStream fos = new FileOutputStream(tempFile)) {
+                fos.write(file.getBytes());
+            }
+
+            File modifiedPDFFile = getModifiedPDFFile(tempFile);
```
spokenbird, line 422: I might be confused, but are we calling `getModifiedPDFFile` before renaming it, thus the file name wouldn't be mutated to include `_modified` yet, causing `getModifiedPDFFile` not to find anything?

spokenbird: Or maybe it's that getModifiedPDFFile should be looking for upload_originalFileName instead of tempFile.getAbsolutePath()_modified.pdf ?

spokenbird (approved):

PR #850, opened 2025-08-25
[CCAP-1259] There's no need to stamp a new PDF

cram-cfa:
#### Issue tracking number 🔗
https://codeforamerica.atlassian.net/browse/CCAP-1259

#### Description of change ✍️
We're not stamping text or annotations, so we can actually just make a copy and avoid this error. The underlying issue is that this PDF has inline values instead of indirect references-- this could happen in the future for other PDFs as well, so this is a worthwhile simple fix.


cram-cfa:
I tested this locally with the PDF that caused the issue in the Jira ticket; I didn't get an error, and I was able to download both the original and converted PDFs from S3 to confirm they both looked correct.

I also used a PDF that needs the settings relaxed, and I confirmed in S3 that both files exist, and the new copy has the settings relaxed.

mtodd-cfa (approved):

spokenbird (approved):
Looks good

